PR #2, opened 2015-08-28
Add hooks for blacklist/whitelist regexs for graphite listener

jbau:
@panghy 

Trying again

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/GraphiteStringHandler.java
+++ b/proxy/src/main/java/com/wavefront/agent/GraphiteStringHandler.java
@@ -3,16 +3,21 @@ package com.wavefront.agent;
 import com.wavefront.agent.api.ForceQueueEnabledAgentAPI;
 import com.wavefront.agent.formatter.Formatter;
 import com.wavefront.ingester.graphite.GraphiteDecoder;
+import com.yammer.metrics.Metrics;
+import com.yammer.metrics.core.Counter;
+import com.yammer.metrics.core.MetricName;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.SimpleChannelInboundHandler;
 import org.apache.commons.lang.StringUtils;
 import sunnylabs.report.ReportPoint;
 
+import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


proxy/src/main/java/com/wavefront/agent/GraphiteStringHandler.java
```diff
@@ -41,13 +55,28 @@ public GraphiteStringHandler(final ForceQueueEnabledAgentAPI agentAPI,
...
     this.formatter = formatter;
+
+    this.pointLineWhiteList = StringUtils.isBlank(pointLineWhiteListRegex) ?
+        null : Pattern.compile(pointLineWhiteListRegex);
+
+    this.pointLineBlackList = StringUtils.isBlank(pointLineBlackListRegex) ?
+        null : Pattern.compile(pointLineBlackListRegex);
+
+    this.regexTimer = Metrics.newTimer(
+        new MetricName("validationRegex." + String.valueOf(port), "", "microsecs"),
```
panghy: I don't think we really need to time regex speeds :)

proxy/src/main/java/com/wavefront/agent/GraphiteStringHandler.java
```diff
@@ -60,6 +89,17 @@ public static String joinPushData(List<String> pushData) {
     return StringUtils.join(pushData, PUSH_DATA_DELIMETER);
   }
 
+  protected void checkWhiteBlackList(String pointLine) {
```
panghy: Return true/false instead of throwing

proxy/src/main/java/com/wavefront/agent/GraphiteStringHandler.java
jbau: I think this is just in case there's something ridiculous (Sam was a bit worried that we couldn't predict the compute expense of the regex that the customer supplies)

(But probably the metric machinery might be as expensive as the regex =P, so I could take it out if you want )

panghy: Yeah, the timer is efficient but not free. If they have a problem, I think we can just ask them to give us the regex and we'd know (probably right away) whether it's a problem or not. Tracking every call seems a bit heavy.

jbau:
repushed

PR #3, opened 2015-09-11
added explicit dependency on jetty-util to fix exception

jumpedoverthemoon:
Caused by: java.lang.ClassNotFoundException: org.eclipse.jetty.util.thread.NonBlockingThread


panghy:
LGTM

PR #4, opened 2015-10-02
Rename blacklist/whitelist conf file params, but maintain compatibility

jbau:
@conorbev  requested a better name


jbau:
please review @panghy   thanks!

panghy:
LGTM

PR #6, opened 2015-11-12
Configure JCommander to accept unparsed flags.

moribellamy:
Need this for a hermetic deployment in which we bundle a JRE.

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
+++ b/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
@@ -1,6 +1,7 @@
 package com.wavefront.agent;
 
 import com.google.common.base.Charsets;
+import com.google.common.base.Joiner;
 import com.google.common.base.Strings;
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
```diff
@@ -240,7 +245,9 @@ private void loadListenerConfigurationFile() throws IOException {
    */
   public void start(String[] args) throws IOException {
     try {
+      logger.info("Arguments: " + Joiner.on(", ").join(args));
       new JCommander(this, args);
+      logger.info("Unparsed arguments: " + Joiner.on(", ").join(unparsed_params));
```
panghy: can unparsed_params be null?

moribellamy: Yes, thanks. FIxed.

panghy:
Please rebase from master

moribellamy:
Done

PR #11, opened 2015-12-08
Faster validations + add validation for point tag keys

jbau:
@panghy 

The validation is definitely faster.  But now there's more =)  Hopefully it will balance out.


jbau:
on synthetic benchmark the validation takes 1/3-1/4 of the time it used to.

PR #18, opened 2016-02-14
Dropwizard metrics

conorbev:
This adds DropWizard Metrics Reporters which use the Wavefront Proxy for the stable version of DropWizard Metrics (which only supports point tags at the Reporter level) and the latest Master of DropWizard Metrics (which supports point tags at the Reporter and Metric level)

It seems like the way JVM metrics are collected has changed quite a bit since the YammerMetrics version that the existing JsonMetricsReporter uses (For instance the VirtualMachineMetrics class no longer exists). 

I have tried to implement something similar via the .includeJvmMetrics() option that can be passed to the Reporter at build time. A user could also ignore that option (it is not on by default) and add exactly whichever JVM metrics they wish to their own MetricsRegistry.

### DIFF
--- a/dropwizard-metrics/3.1/pom.xml
+++ b/dropwizard-metrics/3.1/pom.xml
@@ -9,9 +9,9 @@
     <relativePath>../../pom.xml</relativePath>
   </parent>
   
-  <artifactId>metrics-wavefront</artifactId>
-  <version>3.1.2</version>
-  <name>Wavefront Dropwizard Metrics Reporter</name>
+  <artifactId>dropwizard-metrics-3.1</artifactId>
+  <version>3.7-SNAPSHOT</version>
+  <name>Wavefront Dropwizard Metrics 3.1 Reporter</name>
   <description>Report metrics via the Wavefront Proxy</description>
 
   <dependencies>
--- a/dropwizard-metrics/3.1/pom.xml
+++ b/dropwizard-metrics/3.1/pom.xml
@@ -1,0 +1,35 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.wavefront</groupId>
+    <artifactId>wavefront</artifactId>
+    <version>3.7-SNAPSHOT</version>
+    <relativePath>../../pom.xml</relativePath>
+  </parent>
+  
+  <artifactId>metrics-wavefront</artifactId>
+  <version>3.1.2</version>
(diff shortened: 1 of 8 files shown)


dropwizard-metrics/3.1/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.wavefront</groupId>
+    <artifactId>wavefront</artifactId>
+    <version>3.7-SNAPSHOT</version>
+    <relativePath>../../pom.xml</relativePath>
+  </parent>
+  
+  <artifactId>metrics-wavefront</artifactId>
```
panghy: dropwizard-metrics instead of metrics-wavefront? (it's odd when the package is com.wavefront.metrics-wavefront with two wavefronts)

dropwizard-metrics/3.1/src/main/java/com/wavefront/integrations/metrics/WavefrontReporter.java
```diff
@@ -0,0 +1,378 @@
...
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A reporter which publishes metric values to a Wavefront Proxy.
+ *
+ */
+public class WavefrontReporter extends ScheduledReporter {
+    /**
```
panghy: Indenting is off for this class, we are using a 2 space indenting elsewhere

dropwizard-metrics/3.1/src/main/java/com/wavefront/integrations/metrics/WavefrontReporter.java
```diff
@@ -0,0 +1,378 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(WavefrontReporter.class);
+
+    private final WavefrontSender wavefront;
+    private final Clock clock;
+    private final String prefix;
+    private final String source;
+    private final Map<String, String> pointTags;
+
+    private WavefrontReporter(MetricRegistry registry,
+    						 String proxyHostname,
```
panghy: alignment is off

dropwizard-metrics/4.0/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.wavefront</groupId>
+    <artifactId>wavefront</artifactId>
+    <version>3.7-SNAPSHOT</version>
+    <relativePath>../../pom.xml</relativePath>
+  </parent>
+  
+  <artifactId>metrics-wavefront</artifactId>
```
panghy: dropwizard-metrics-4.0

conorbev: OK, I'll switch to that for artifact idea and change the version to 3.7-SNAPSHOT to match everything else in the project. This way it'll still remain clear that this version is for use with DropWizard Metrics 4.0 while still allowing us to have an independent version number for the Wavefront Reporter that we can increment as we add features and fix bugs.

PR #19, opened 2016-02-15
Configureable custom source tags

conorbev:
We already had a hard coded "extra" tag that could resolve to source/host which was "fqdn".

This adds the ability for a user to specify additional tags that should be used as the source in cases where neither a source= or host= tag exists

### DIFF
--- a/java-lib/src/main/java/com/wavefront/ingester/GraphiteDecoder.java
+++ b/java-lib/src/main/java/com/wavefront/ingester/GraphiteDecoder.java
@@ -5,6 +5,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
 import com.google.common.collect.Lists;
 
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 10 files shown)


java-lib/src/main/java/com/wavefront/ingester/GraphiteDecoder.java
```diff
@@ -26,19 +27,24 @@
       .appendOptionalTimestamp().whiteSpace()
       .appendAnnotationsConsumer().whiteSpace().build();
   private final String hostName;
+  private LinkedHashSet<String> customSourceTags;
 
-  public GraphiteDecoder() {
+  public GraphiteDecoder(LinkedHashSet<String> customSourceTags) {
```
panghy: Why a LinkedHashSet?

panghy: Ah, I see, you need ordering, it's probably better to accept a Set<String> here and pass in an ordered set for use cases that need ordering

java-lib/src/test/java/com/wavefront/ingester/GraphiteDecoderTest.java
```diff
@@ -18,9 +19,11 @@
  */
 public class GraphiteDecoderTest {
 
+  private final static LinkedHashSet<String> emptyCustomSourceTags = new LinkedHashSet<String>();
```
panghy: That way, you can just use Collections.emptySet() here

java-lib/src/main/java/com/wavefront/ingester/GraphiteDecoder.java
conorbev: Alright, I deliberately used the LinkedHashSet here to ensure the ordering would always work as expected, because I think having it ordered is the only use case that ever makes sense here. I wanted to make sure that even if someone was changing what the PushAgent passes down the road they wouldn't wonder "Why LinkedHashSet, I'll just use HashSet" and accidentally break it.

If you're sure I can certainly change it though. Do you want it changed to Set in all arg lists, like in IngesterFormatter.drive() as well ?

panghy: Typically, the idea is to make inputs as generic as possible (since it's perfectly possible that there are other ordered set implementations). I think in your case, you can actually use Collection<String> even and state that the iteration order is the order in which the tag would be considered. You can then even use a list (since you really want an error if someone puts a tag twice anyways in the config/arg).

conorbev: Alright, would you be OK with me switching it to a List? I think the intent of what should happen is then much clearer than Collection with a comment about iteration order mattering.

I only picked LinkedHashSet because I wanted it to attempt to dtrt if a user had repeated stuff in their wavefront.conf.

I think you are right though. It is better to give a warning if someone puts a tag twice than silently fixing it as it probably indicates they should check their config. I think List is therefore better than Set. 

For now I have made it a warning and the program continues but can change that if you like.

PR #20, opened 2016-02-19
Improve UX when single agent is used for high load.

moribellamy:
Also fix proxy/pom.xml so that intellij cleanly builds this project.

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
+++ b/proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
@@ -24,6 +24,7 @@ import com.yammer.metrics.core.Gauge;
 import com.yammer.metrics.core.Histogram;
 import com.yammer.metrics.core.Meter;
 import com.yammer.metrics.core.MetricsRegistry;
+import org.apache.commons.lang.exception.ExceptionUtils;
 
 import javax.annotation.Nullable;
 import javax.ws.rs.core.Response;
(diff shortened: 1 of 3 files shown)


proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
```diff
@@ -430,10 +431,18 @@ private void addTaskToSmallestQueue(ResubmissionTask taskToRetry) {
     TaskQueue<ResubmissionTask> queue = getSmallestQueue();
     if (queue != null) {
       synchronized (queue) {
-        queue.add(taskToRetry);
+        try {
+          queue.add(taskToRetry);
+        } catch (FileException e) {
+          logger.warning(
+              "CRITICAL (Losing points!): " +
+                  "WF-1: Could not open enough file descriptors to send points to " +
```
panghy: This is not the right message, this is actually it running out of queue space and not FDs. We have another way to detect this by looking at the actual FD usage so I think we should just say that we can't write to the queue any more (due to space) and we should definitely add the exception to the message as well

proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
```diff
@@ -430,10 +432,16 @@ private void addTaskToSmallestQueue(ResubmissionTask taskToRetry) {
     TaskQueue<ResubmissionTask> queue = getSmallestQueue();
     if (queue != null) {
       synchronized (queue) {
-        queue.add(taskToRetry);
+        try {
+          queue.add(taskToRetry);
+        } catch (FileException e) {
+          logger.warning(
```
panghy: You can use logger.log(Level.WARN, "...", e);

PR #25, opened 2016-03-22
Enforce 100 char line limit.

moribellamy:


panghy:
I actually checked in the IntelliJ code style, just do another cursory check and we can pull this in

PR #27, opened 2016-03-29
add support for write_http JSON format

wavefront-mike:
This was added to allow customers with older versions of collectd (without write_graphite or write_tsdb).  write_http was added in 4.8 and this allows customers who are on 4.8 or later to send stuff to our proxy via write_http.
See https://collectd.org/wiki/index.php/Plugin:Write_HTTP
Also added a test runner to make sure json is parsed correctly.  We'll need to add some more test case. It is currently testing a very simple case to catch the basics.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ proxy/unitTest*
 .classpath
 .project
 .settings/
+
+test/buffer.*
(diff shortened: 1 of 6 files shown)


panghy:
The branch is not building right now

PR #28, opened 2016-03-30
1) Daemonize the WF proxy, so it will restart on unclean exits. To do

moribellamy:
this we include the Apache Commons Daemon in our package. This can be
compiled on CentOS6 and shipped to many platforms, since CentOS6 has and
old version of glibc. A dockerfile for building the packages is
included.

2) Add proxy to startup scripts so that it loads automatically on
node boot. We continue to use the SysV init methodology because of its
widespread support -- it is supported on both RHEL and Ubuntu, while the
next-gen init systems (upstart et. al) only work on their corresponding
distro.


proxy/pom.xml
```diff
@@ -107,6 +102,16 @@
...
     </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-all</artifactId>
+      <version>4.0.33.Final</version>
+    </dependency>
+    <dependency>
+      <groupId>commons-daemon</groupId>
+      <artifactId>commons-daemon</artifactId>
+      <version>1.0.15</version>
```
panghy: should we have the commons-daemon github not as a submodule but rather as a required dependency to build the project? We obviously depend on a bunch of tools as well as a compiled jdk, can we just assume you have some version of the commons-daemon native code in a local directory? (with some readme that details all the necessary dependencies and tooling needed to build the agent in a readme file of course)

moribellamy: Done -- since we're using a Dockerfile the build is fairly straightforward no matter how we decide to organize things. PTAL.

panghy:
Your last commit doesn't seem to reflect the actual change (I can't see how that's using 600 permissions on the conf file. You can amend that commit and we should be good

PR #30, opened 2016-04-11
Verify point length before sending to WF.

moribellamy:


proxy/src/main/java/com/wavefront/agent/ChannelStringHandler.java
```diff
@@ -136,7 +164,12 @@ protected void channelRead0(ChannelHandlerContext ctx, String msg) throws Except
     }
     if (!points.isEmpty()) {
       ReportPoint point = points.get(0);
-      pointHandler.reportPoint(point, pointLine);
+      String errorMessage = verifyPoint(point);
+      if (errorMessage != "") {
```
panghy: if you don't have intellij, let's fix that but you can't do this in java, you need equals() (intellij will flag that)

proxy/src/main/java/com/wavefront/agent/ChannelStringHandler.java
```diff
@@ -101,6 +102,33 @@ protected boolean passesWhiteAndBlackLists(String pointLine) {
...
+   */
+  private String verifyPoint(ReportPoint point) {
+    StringBuilder sb = new StringBuilder();
+    if (point.getHost().length() >= 1024) {
+      sb.append(" Host too long: ").append(point.getHost()).append(".");
+    }
+    if (point.getMetric().length() >= 1024) {
+      sb.append(" Metric too long: ").append(point.getMetric()).append(".");
+    }
+    // Each tag of the form "k=v" must be < 1024
```
panghy: 256

proxy/src/main/java/com/wavefront/agent/ChannelStringHandler.java
```diff
@@ -101,6 +102,33 @@ protected boolean passesWhiteAndBlackLists(String pointLine) {
...
+    for (Map.Entry<String, String> tag : point.getAnnotations().entrySet()) {
+      if (tag.getKey().length() + tag.getValue().length() >= 1023) {
```
panghy: 256

PR #35, opened 2016-04-18
Support ephemeral agents in java-lib.

moribellamy:
Backwards-compatible to older WF versions -- the "ephemeral" field is
ignored in that case.


proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
```diff
@@ -176,6 +176,9 @@
   @Parameter(names = {"--customSourceTags"}, description = "Comma separated list of point tag keys that should be treated as the source in Wavefront in the absence of a tag named source or host")
   protected String customSourceTagsProperty = "fqdn";
 
+  @Parameter(names = {"--ephemeral"}, description = "If true, this agent is removed from Wavefront after 24 hours of inactivity.")
+  protected boolean ephemeral = false;
```
panghy: You need to set this value from wavefront.conf

moribellamy: Done. I followed the existing pattern in this app (allow CLI, but prefer config file).

PR #45, opened 2016-05-25
Support OpenTSDB telnet and HTTP protocols on a single port

wavefront-mike:
## Background Information
OpenTSDB actually listens for the "telnet" protocol and the HTTP protocol on a single port (4242).  It detects the data type and handles the command based on the request.
See [OpenTSDB documentation](http://opentsdb.net/docs/build/html/user_guide/writing.html#input-methods) for more details.

## Overview of Changes
- This pull request adds a netty decoder similar to the one implemented in the tsd code [https://github.com/OpenTSDB/opentsdb/blob/master/src/tsd/PipelineFactory.java].  It uses the first 2 bytes to detect if the request is HTTP or telnet.  The paths for the HTTP request are checked manually instead of with JAX-RS because the [connector](https://github.com/caskdata/netty-http) to netty does not seem not very pluggable.

- The previously unsupported OpenTSDB "version" command is now supported on both HTTP and telnet.  tcollector requires this command to work properly with the proxy.  Without this command, the proxy will get [blacklisted](https://github.com/OpenTSDB/tcollector/blob/master/tcollector.py#L533) by tcollector.

## Major changes in this pull request:
- Ingester was split into multiple classes: TcpIngester, StringLineIngester
- Moved common code for metrics to PointHandler from ChannelStringHandler (including prefix management and metric/host/annotation validation) so it could be re-used
- added PlainTextOrHttpFrameDecoder class to facilitate switching between plain text and HTTP protocols
- PushAgent.startOpenTsdbListener() now uses the OpenTSDBPortUnificationHandler class
- pylint run on test/runner.py and more tests added to automate testing most OpenTSDB code

## Tests
See test/runner.py to run automated tests.  This script starts up a fake HTTP endpoint for API calls and then runs tests against most of the supported protocols/ports.  It then confirms that the HTTP endpoint received metrics as expected.

### DIFF
--- a/java-lib/src/main/java/com/wavefront/ingester/Ingester.java
+++ b/java-lib/src/main/java/com/wavefront/ingester/Ingester.java
@@ -37,8 +37,7 @@ public abstract class Ingester implements Runnable {
   private static final Logger logger = Logger.getLogger(Ingester.class.getCanonicalName());
 
   /**
-   * Default number of seconds before the channel idle timeout handler
-   * closes the connection.
+   * Default number of seconds before the channel idle timeout handler closes the connection.
    */
   private static final int CHANNEL_IDLE_TIMEOUT_IN_SECS_DEFAULT = (int) TimeUnit.DAYS.toSeconds(1);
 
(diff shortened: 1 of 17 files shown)


wavefront-mike:
Going to have @ezeev and @conorbev take a look at this pull request first ...

conorbev:
Looks good to me. I don't think it affects this PR directly but could you please work with @durren to get his python test suite merged with yours?

java-lib/src/main/java/com/wavefront/ingester/Ingester.java
```diff
@@ -31,76 +33,92 @@
...
-      // Server was interrupted
-      e.printStackTrace();
+  /**
+   * Adds an idle timeout handler to the given pipeline
+   * @param pipeline the pipeline to add the idle timeout handler
+   */
+  protected void addIdleTimeoutHandler(final ChannelPipeline pipeline) {
+    // Shared across all reports for proper batching
+    pipeline.addLast("idleStateHandler",
+                     new IdleStateHandler(CHANNEL_IDLE_TIMEOUT_IN_SECS_DEFAULT,
```
panghy: method invocations do not align like this, check your IDE settings (they should just be 4-character indented)

java-lib/src/main/java/com/wavefront/ingester/Ingester.java
```diff
@@ -31,76 +33,92 @@
...
+                                          0, 0));
+    pipeline.addLast("idleChannelTerminator", new ChannelDuplexHandler() {
+        @Override
+        public void userEventTriggered(ChannelHandlerContext ctx,
```
panghy: same here for method declaration line continuation

java-lib/src/main/java/com/wavefront/ingester/StringLineIngester.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  /**
+   * Returns a copy of the given list plus inserts the 2 decoders needed for this specific ingester
+   * (LineBasedFrameDecoder and StringDecoder)
+   *
+   * @param decoders the starting list
+   * @return copy of the provided list with additional decodiers prepended
+   */
+  private static List<Function<Channel, ChannelHandler>> createDecoderList(final List<Function<Channel, ChannelHandler>> decoders) {
```
panghy: decoders need to mark @Nullable

proxy/src/main/java/com/wavefront/agent/PlainTextOrHttpFrameDecoder.java
```diff
@@ -0,0 +1,108 @@
...
+import io.netty.handler.codec.http.HttpContentCompressor;
+import io.netty.handler.codec.http.HttpObjectAggregator;
+import io.netty.handler.codec.http.HttpRequestDecoder;
+import io.netty.handler.codec.http.HttpResponseEncoder;
+
+import java.util.List;
+import java.util.logging.Logger;
+
+/**
+ * This class handles 2 d ifferent protocols on a single port.  Supported protocols include HTTP and
```
panghy: different

PR #49, opened 2016-06-15
Data Dog Agent JSON Handler

wavefront-mike:
Handler for Data Dog's JSON data


wavefront-mike:
switching to use custom emitter instead

PR #55, opened 2016-07-07
Patch httpClient with HTTPCLIENT-1747 fix

basilisk487:


panghy:
Can you add comments to the pom.xml and a README to the patched directory to indicate when we can remove this?

basilisk487:
Added the unit test to verify that the patches work + README and pom.xml comments

PR #62, opened 2016-07-11
Workaround for JDK bug JDK-8075484

basilisk487:


panghy:
This won't work in the long run, it'll just leak threads when the issue happens

PR #79, opened 2016-08-10
Graceful service shutdown and configure SO_LINGER

basilisk487:
Prevent point loss during proxy shutdown/restart process - stop port listeners first and drain all in-memory buffers to retry queue

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/PushAgent.java
+++ b/proxy/src/main/java/com/wavefront/agent/PushAgent.java
@@ -26,6 +26,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
+import java.util.IdentityHashMap;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -274,4 +278,29 @@ protected void processConfiguration(AgentConfiguration config) {
...
+    managedThreads.add(thread);
+    thread.start();
+  }
+
+  @Override
+  public void stopListeners() {
+    for (Thread thread : managedThreads) {
+      thread.interrupt();
+      int i = 0;
+      while (i < TimeUnit.SECONDS.toMillis(10) / 50) { // give up to 10 seconds to stop a listener thread
```
panghy: why not Thread.join() with a timeout?

basilisk487:
@panghy Updated as per code review + added ability to configure SO_LINGER on listener ports as per customer's request

PR #80, opened 2016-08-15
Reduce memory footprint

basilisk487:
- No longer convert every single point to a string just in case it throws an exception
- When splitting a batch, use a string index instead of recreating all the strings

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/PointHandler.java
+++ b/proxy/src/main/java/com/wavefront/agent/PointHandler.java
@@ -2,6 +2,8 @@ package com.wavefront.agent;
 
 import java.util.List;
 
+import javax.annotation.Nullable;
+
 import sunnylabs.report.ReportPoint;
 
 /**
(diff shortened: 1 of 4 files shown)


panghy:
Have you tested that it does save memory? I believe String substrings actually create a copy of the char[] array instead of sharing. An allocation or garbage trace would show that.

basilisk487:
@panghy It does extend full GC cycles a bit and gives 5-7% reduction in both PS_Scavenge and PS_MarkSweep GC runs. In addition, it boosts the proxy's throughput by about 9% simply by not having to run a point to string conversion twice.

proxy/src/main/java/com/wavefront/agent/PointHandler.java
```diff
@@ -14,9 +16,9 @@
    * Send a point for reporting.
    *
    * @param point     Point to report.
-   * @param debugLine Debug information to print to console when the line is rejected.
+   * @param debugLine Debug information to print to console when the line is rejected. If null, then
```
panghy: If null, then?

panghy:
I guess does the logic in QueuedAgentService actually helps performance though or is it just the savings on pointToString() that's helping?

basilisk487:
@panghy The logic in QueuedAgentService doesn't affect performance noticeably, but it does help  to reduce memory use when processing the retry queue under heavy load. By the time it gets around to process the batch from the queue, the original strings representing separate points will have been garbage collected (probably long time ago too) - and the new algorithm doesn't recreate the original strings when splitting the task so it uses half the memory per split batch.

panghy:
ok, it would seem to me that the actual impact should not be that significant, 5-7% isn't much (and GC is somewhat random). anyhow, will merge

PR #81, opened 2016-08-17
Prevent OoM in a tight retry loop 

basilisk487:
on a machine with very limited resources. 
When the number of push threads is low and there are a lot of items in the retry queue; when a  new batch is also forced to be queued, the retry loop is very likely to immediately acquire back the just released lock, not giving the push thread a chance to write to the queue - and this could last for several minutes, quickly eating up heap memory as points are accumulating in a push thread's buffers.

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
+++ b/proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
@@ -221,6 +221,7 @@ public class QueuedAgentService implements ForceQueueEnabledAgentAPI {
                 } finally {
                   if (removeTask) taskQueue.remove();
                 }
+                taskQueue.wait(50); // release lock for 50ms  
               }
             }
           } catch (Throwable ex) {
--- a/proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
+++ b/proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
@@ -169,8 +169,8 @@ public class QueuedAgentService implements ForceQueueEnabledAgentAPI {
               logger.warning("[RETRY THREAD " + threadId + "] TASK STARTING");
             }
             while (taskQueue.size() > 0 && taskQueue.size() > failures) {
+              taskQueue.getLockObject().lock();
               try {
-                taskQueue.getLockObject().lock();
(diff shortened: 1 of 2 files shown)


basilisk487:
@panghy: Switched from waiting for 50ms to using a less crude approach by using a ReentrantLock in "fair" mode. It also solves the thread starvation issue:
Before:
image

After:
image

proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
```diff
@@ -451,15 +454,18 @@ public Response postPushData(UUID agentId, UUID workUnitId, Long currentMillis,
   }
 
   private void addTaskToSmallestQueue(ResubmissionTask taskToRetry) {
-    TaskQueue<ResubmissionTask> queue = getSmallestQueue();
+    ResubmissionTaskQueue queue = getSmallestQueue();
     if (queue != null) {
-      synchronized (queue) {
+      try {
```
panghy: these two try blocks can be collapsed

panghy: also, it's normally lock then try ... finally {unlock} (not that it matters i believe in practice but let's say it OOM right after you enter the try block...)

panghy: on second thought, i guess putting it outside the loop can also lock without releasing (depending on how the compiler weaves the code together), anyhow, it's just convention to lock then try :)

proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
```diff
@@ -169,7 +169,8 @@ public void run() {
               logger.warning("[RETRY THREAD " + threadId + "] TASK STARTING");
             }
             while (taskQueue.size() > 0 && taskQueue.size() > failures) {
-              synchronized (taskQueue) {
+              try {
+                taskQueue.getLockObject().lock();
```
panghy: ditto here for lock then try

panghy:
Thanks for looking into this

basilisk487:
@panghy Done!

PR #84, opened 2016-08-29
Fall back to using an IP address if RDNS lookup fails

basilisk487:


proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
```diff
@@ -693,4 +693,29 @@ public void shutdown() {
...
+        // if can't resolve local server name, use a real IPv4 address from the first available network interface
+        for (Enumeration<NetworkInterface> nics = NetworkInterface.getNetworkInterfaces();
+             nics.hasMoreElements();) {
+          NetworkInterface network = nics.nextElement();
+          if (!network.isUp() || network.isLoopback())
+            continue;
+          for (Enumeration<InetAddress> addresses = network.getInetAddresses(); addresses.hasMoreElements();) {
+            InetAddress address = addresses.nextElement();
+            if (address.isAnyLocalAddress() || address.isLoopbackAddress() ||
+                address.isMulticastAddress() || !(address instanceof Inet4Address))
```
panghy: don't like ipv6 only machines? :)

PR #91, opened 2016-09-15
More flexible point filtering and altering at the proxy 

basilisk487:
Allow arbitrary replacements (regex) on the whole point line as well as on individual point components (metric name, source name, point tag) after parsing. Streamline the whole approach to avoid constantly changing signature for multiple methods when passing arguments through multiple classes. "Metric name prefix" and "blacklistRegex/whitelistRegex" features ported to the new framework for complete backwards compatibility.


proxy/src/test/java/com/wavefront/agent/PreprocessorRulesTest.java
```diff
@@ -0,0 +1,274 @@
...
+
+  @BeforeClass
+  public static void setup() throws IOException {
+    InputStream stream = PreprocessorRulesTest.class.getResourceAsStream("preprocessor_rules.yaml");
+    config = new AgentPreprocessorConfiguration();
+    config.loadFromStream(stream);
+  }
+
+  @Test(expected = NullPointerException.class)
+  public void testInvalidPointLineRule1() {
```
moribellamy: if you want to have one test case for each invalid argument, i'd give each test case a more descriptive name, e.g.

> testNullRegexReplaceRuleThrowsFor...

If you don't care about such descriptiveness, I'd put all of the test cases in the same method, and use something like http://blog.codeleak.pl/2015/04/junit-testing-exceptions-with-java-8.html or https://code.google.com/archive/p/catch-exception/

basilisk487: Fixed!

proxy/src/test/java/com/wavefront/agent/PreprocessorRulesTest.java
```diff
@@ -0,0 +1,274 @@
...
+    assertEquals(expectedPoint2, rule2.apply(testPoint2));
+    assertTrue(rule3.apply(testPoint1));
+    assertFalse(rule3.apply(testPoint2));
+    assertFalse(rule4.apply(testPoint1));
+    assertTrue(rule4.apply(testPoint2));
+  }
+
+  @Test
+  public void testReportPointRules() {
+    Map<String, String> annotations = new TreeMap<>();
```
basilisk487: Fixed - went a different route (parse the point from a string instead of creating it on the fly)

proxy/src/main/java/com/wavefront/agent/preprocessor/PointLineReplaceRegexTransformer.java
```diff
@@ -0,0 +1,48 @@
...
+ * Created by Vasily on 9/13/16.
+ */
+public class PointLineReplaceRegexTransformer implements Function<String, String> {
+
+  private final String patternReplace;
+  private final Pattern compiledPattern;
+  @Nullable
+  private final Counter ruleAppliedCounter;
+
+  public PointLineReplaceRegexTransformer(final String patternMatch,
```
basilisk487: Probably not here - @Notnull annotation only helps the IDE to give a soft warning if a null value can potentially sneak in (and here it obviously can, as it's being parsed from yaml - hence the validation in the constructor)

proxy/src/main/java/com/wavefront/agent/preprocessor/PointLineReplaceRegexTransformer.java
```diff
@@ -0,0 +1,48 @@
...
+                                          final String patternReplace,
+                                          @Nullable final Counter ruleAppliedCounter) {
+    Preconditions.checkNotNull(patternMatch, "[match] can't be null");
```
basilisk487: Fixed (where it made sense)

proxy/src/main/java/com/wavefront/agent/ChannelByteArrayHandler.java
```diff
@@ -63,9 +56,17 @@ protected void channelRead0(ChannelHandlerContext ctx, byte[] msg) throws Except
...
-        if (!this.metricPredicate.apply(point.getMetric())) {
-          pointHandler.handleBlockedPoint(point.getMetric());
-          continue;
+        if (preprocessor != null) {
+          // backwards compatibility: apply "pointLine" rules to metric name
+          if (!preprocessor.forPointLine().filter(point.getMetric())) {
+            pointHandler.handleBlockedPoint(preprocessor.forReportPoint().getLastFilterResult());
+            continue;
+          }
+          preprocessor.forReportPoint().transform(point);
```
moribellamy: it looks like this block of code is the entry point to all stream processing you've added in this PR, is that correct?

proxy/src/main/java/com/wavefront/agent/preprocessor/AgentPreprocessorConfiguration.java
```diff
@@ -0,0 +1,162 @@
...
+      //noinspection unchecked
+      Map<String, Object> rulesByPort = (Map<String, Object>) yaml.load(stream);
+      for (String strPort : rulesByPort.keySet()) {
+        int validRules = 0;
+        //noinspection unchecked
+        List<Map<String, String>> rules = (List<Map<String, String>>) rulesByPort.get(strPort);
+        for (Map<String, String> rule : rules) {
+          try {
+            requireArguments(rule, "rule", "action");
+            allowArguments(rule, "rule", "action", "scope", "search", "replace", "match", "tag", "newtag", "value");
```
moribellamy: very nice fail fast

proxy/src/main/java/com/wavefront/agent/ChannelByteArrayHandler.java
basilisk487: This, and https://github.com/wavefrontHQ/java/pull/91/files/b1a67233b867796c5177ea9c5c8c03bfba610f59#diff-218e7fd1a369e3a4a715cc0ee4405ed4L90

PR #96, opened 2016-09-30
Support for direct logs ingestion from filebeat -> WF proxy.

moribellamy:
Initial version has support for counters and Gauges.

To use this new feature:

* in your config file (wavefront.conf), specify `logsIngestionConfigFile=logsIngestion.yaml`
* logsIngestion.yaml has a configuration object, namely LogsIngestionConfig. for example, you could use the following config to send a counter to wavefront:

```
 - pattern: "operation %{WORD:op} took %{NUMBER:value} seconds in DC=%{WORD:dc}.*AZ=%{WORD:az}"
    metricName: "%{op}.totalSeconds"
    tagKeys:
      - "theDC"
      - "theAZ"
    tagValueLabels:
      - "dc"
      - "az"
```

so, if you had a log line that was "operation foo took 2 seconds in DC=wavefront AZ=2a”, we would notice a match in your config, and construct the point  `foo.totalSeconds 2 theDC=wavefront theAZ=2a`

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/PushAgent.java
+++ b/proxy/src/main/java/com/wavefront/agent/PushAgent.java
@@ -6,6 +6,7 @@ import com.google.common.base.Splitter;
 
 import com.beust.jcommander.internal.Lists;
 import com.wavefront.agent.formatter.GraphiteFormatter;
+import com.wavefront.agent.logsharvesting.FilebeatListener;
 import com.wavefront.agent.preprocessor.PointPreprocessor;
 import com.wavefront.agent.preprocessor.ReportPointAddPrefixTransformer;
 import com.wavefront.agent.preprocessor.ReportPointTimestampInRangeFilter;
@@ -22,6 +23,7 @@ import com.wavefront.ingester.TcpIngester;
 import org.glassfish.jersey.jackson.JacksonFeature;
 import org.glassfish.jersey.jetty.JettyHttpContainerFactory;
 import org.glassfish.jersey.server.ResourceConfig;
+import org.logstash.beats.Server;
 
 import java.io.IOException;
 import java.net.URI;
@@ -152,6 +154,20 @@ public class PushAgent extends AbstractAgent {
         }
       }
     }
+
+    if (filebeatPort != null) {
+      final Server filebeatServer = new Server(filebeatPort);
+      filebeatServer.setMessageListener(new FilebeatListener(
+          new PointHandlerImpl(filebeatPort, pushValidationLevel, pushBlockedSamples, getFlushTasks(filebeatPort)),
+          logsIngestionConfig, hostname, prefix, true));
+      startAsManagedThread(() -> {
+        try {
+          filebeatServer.listen();
+        } catch (InterruptedException e) {
+          e.printStackTrace();
+        }
+      });
+    }
   }
 
   protected void startOpenTsdbListener(final String strPort) {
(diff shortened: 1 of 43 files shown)


proxy/src/main/java/com/wavefront/agent/logsharvesting/FilebeatListener.java
```diff
@@ -0,0 +1,185 @@
...
+   */
+  public FilebeatListener(PointHandler pointHandler, LogsIngestionConfig logsIngestionConfig, String hostname,
+                          String prefix, boolean reportMetaMetrics) {
+    this.pointHandler = pointHandler;
+    this.hostname = hostname;
+    this.prefix = prefix;
+    this.logsIngestionConfig = logsIngestionConfig;
+    this.metricsRegistry = new MetricsRegistry();
+    // Meta metrics.
+    this.received = metricsRegistry.newCounter(FilebeatListener.class, "received");
```
basilisk487: Is there a reason why these metrics are sent outside of the usual "agent metrics" flow? There's also a possibility that it might prevent them from being replicated to 'mon' instance to be monitored centrally.

`this.received = Metrics.newCounter(new MetricName("logsharvesting", "", "received"));`
would automatically register this metric and it will be sent along with the rest during agent check-in, so a separate scheduled task would be not necessary. 
If we do need to send these particular metrics more frequently (like every second), should we move them to a different namespace?

moribellamy: done

proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -155,11 +155,11 @@ protected void startListeners() {
       }
     }
 
-    if (filebeatPort != null) {
+    if (filebeatPort > 0) {
       final Server filebeatServer = new Server(filebeatPort);
       filebeatServer.setMessageListener(new FilebeatListener(
           new PointHandlerImpl(filebeatPort, pushValidationLevel, pushBlockedSamples, getFlushTasks(filebeatPort)),
-          logsIngestionConfig, hostname, prefix, true));
+          logsIngestionConfig, hostname, prefix));
```
basilisk487: if filebeatPort is > 0 and logsIngestionConfig is null, it will crash with NPE at start-up and from the exception text it's not obvious how can it be fixed - so it might be a good idea to check if both variables are set in the config file

proxy/src/main/java/com/wavefront/agent/logsharvesting/FilebeatListener.java
```diff
@@ -47,41 +48,20 @@
...
-              .setHost(hostname).setValue(unparsed.count()).setTimestamp(System.currentTimeMillis()).build(), null);
-          pointHandler.reportPoint(ReportPoint.newBuilder().setMetric("~agent.logsharvesting.parsed")
-              .setHost(hostname).setValue(parsed.count()).setTimestamp(System.currentTimeMillis()).build(), null);
-          pointHandler.reportPoint(ReportPoint.newBuilder().setMetric("~agent.logsharvesting.batchesProcessed")
-              .setHost(hostname).setValue(batchesProcessed.count())
-              .setTimestamp(System.currentTimeMillis()).build(), null);
-        }
-      }, 1000, 1000);
-    }
+    this.received = Metrics.newCounter(FilebeatListener.class, "logsharvesting.received");
```
basilisk487: Almost there :) 
However, using FilebeatListener.class would end up generating a really long metric name (`~agent.com.wavefront.agent.logsharvesting.logsharvesting.received`) which is probably not what was intended. The workaround is to create a metric name under a specific "namespace" (which is actually 2nd level namespace): `new MetricName("logsharvesting", "", "received")`

moribellamy: done

PR #105, opened 2016-10-24
Limit rate at the proxy

basilisk487:
Limit outgoing rate at the proxy (averaged over 1 minute), with a single limit controlling both data flows (incoming points and retry queue). Incoming points are always given a priority, and retry queue is considered ok to start processing if there is more than 1 second worth of accumulated unused "credits" (which means that the rate for incoming points is less than the limit).

### DIFF
--- a/java-lib/src/main/java/com/wavefront/ingester/StringLineIngester.java
+++ b/java-lib/src/main/java/com/wavefront/ingester/StringLineIngester.java
@@ -84,4 +84,12 @@ public class StringLineIngester extends TcpIngester {
     index.add(pushData.length());
     return index;
   }
+
+  public static int pushDataSize(String pushData) {
+    int length = StringUtils.countMatches(pushData, PUSH_DATA_DELIMETER);
+    return length > 0
+        ? length + 1
+        : (pushData.length() > 0 ? 1 : 0);
+
+  }
 }
--- a/java-lib/src/main/java/com/wavefront/ingester/StringLineIngester.java
+++ b/java-lib/src/main/java/com/wavefront/ingester/StringLineIngester.java
@@ -85,6 +85,11 @@ public class StringLineIngester extends TcpIngester {
     return index;
   }
 
+  /**
+   * Calculates the number of points in the pushData payload
+   * @param pushData a delimited string with the points payload
+   * @return number of points
+   */
   public static int pushDataSize(String pushData) {
     int length = StringUtils.countMatches(pushData, PUSH_DATA_DELIMETER);
     return length > 0
--- a/proxy/src/main/java/com/google/common/util/concurrent/RecyclableRateLimiter.java
+++ b/proxy/src/main/java/com/google/common/util/concurrent/RecyclableRateLimiter.java
@@ -59,6 +59,11 @@ public class RecyclableRateLimiter extends RateLimiter {
     this.setRate(permitsPerSecond);
   }
 
+  /**
+   * Get the number of accumulated permits
+   *
+   * @return number of accumulated permits
+   */
   public double getAvailablePermits() {
     synchronized (mutex) {
       resync(stopwatch.readMicros());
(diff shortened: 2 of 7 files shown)


java-lib/src/main/java/com/wavefront/ingester/StringLineIngester.java
```diff
@@ -84,4 +84,12 @@ public static String joinPushData(List<String> pushData) {
     index.add(pushData.length());
     return index;
   }
+
+  public static int pushDataSize(String pushData) {
+    int length = StringUtils.countMatches(pushData, PUSH_DATA_DELIMETER);
```
moribellamy: what if there is an escaped delimeter in the payload?

proxy/src/main/java/com/google/common/util/concurrent/RecyclableRateLimiter.java
```diff
@@ -0,0 +1,138 @@
+package com.google.common.util.concurrent;
+
+import com.google.common.math.LongMath;
+
+import static java.lang.Math.max;
+import static java.lang.Math.min;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * An alternative RateLimiter implementation that allows to "return" unused permits back to the pool to handle retries
```
moribellamy: Is it possible to use a normal guava rate limiter, but to continually modify its rate? That would be easier to review at least :), and less error prone.

proxy/src/main/java/com/wavefront/agent/PostPushDataTimedTask.java
```diff
@@ -122,30 +138,60 @@ public UUID getDaemonId() {
...
-      batchesSent.inc();
+      int batchSize;
+      List<String> current = null;
+      batchesAttempted.inc();
+      numIntervals++;
+
+      synchronized (pointsMutex) {
+        batchSize = Math.min(points.size(), pointsPerBatch);
+        if (batchSize != 0) {
+          if (!pushRateLimiter.tryAcquire(batchSize)) {
```
moribellamy: acquiring a permit while holding the pointsMutex lock seems fishy. can you comment in your reasoning here?

java-lib/src/main/java/com/wavefront/ingester/StringLineIngester.java
basilisk487: Strictly speaking, that should never happen as a delimiter sequence would be considered invalid characters. StringLineIngester.joinPushData and unjoinPushData are also operating under an assumption that PUSH_DATA_DELIMITER doesn't naturally occur in the payload.

proxy/src/main/java/com/wavefront/agent/PostPushDataTimedTask.java
basilisk487: tryAcquire returns "false" immediately if no permits are available, so there is no delay introduced.  The thinking was to avoid having to splice the batch of points back into the beginning of the queue (to retain the in-memory point order), if no permits are available:  tryAcquire while holding lock should still be faster than `current = points.subList(0, blockSize); points = new ArrayList<>(points.subList(blockSize, points.size()));` in createAgentPostBatch(), and then `points.addAll(0, current);`, with both of these holding the pointsMutex lock.

proxy/src/main/java/com/google/common/util/concurrent/RecyclableRateLimiter.java
basilisk487: Unfortunately, standard guava rate limiter has a couple of limitations:
- it doesn't allow accumulating "credits" over a time period other than 1 second - the SmoothBursty constructor where you can specify maxBurstSeconds is package-private, and   RateLimiter.create invokes the the constructor with a hardcoded value of "1.0" for maxBurstSeconds. The only way around it is to have a complete implementation.
- it doesn't allow querying the status and seeing if / how many permits are currently available. Having this functionality makes it possible to have only one rate limit. The alternative would be to have separate rate limits for incoming points and for the retry queue, but users don't care (and shouldn't care) about the difference, as they just want to control the rate of points leaving the proxy. With separate rate limiters in place slightly different circumstances can produce vastly different results, so the goal was to have a single limiter.

moribellamy:
Sorry, please fix the conflicts and i'll take another look.

Also, please make sure each method is either annotated Override, or is private, or has docstring.

basilisk487:
@moribellamy: Done!

PR #111, opened 2016-10-27
Improve preprocessor rules

basilisk487:
- Allow pointLine scope rules for binary protocols like graphite pickle, by faking "plaintext" protocol by converting a point to a string and parsing it back. The goal is to provide temporary implementation for the same set of rules during POC trials, when Graphite and Wavefront consume the same data in parallel from carbon-relay, which can only output data in the binary pickle format. Once the trial is over, carbon-relay can be removed from the pipeline and switch back to plaintext protocol.
- Add an optional "match" condition to replaceRegex rules: when defined, apply search/replace regex only if the string matches. In some cases this can significantly simplify search/replace expressions.

### DIFF
--- a/proxy/pom.xml
+++ b/proxy/pom.xml
@@ -87,16 +87,6 @@
       <version>3.1</version>
     </dependency>
     <dependency>
-      <groupId>com.squareup</groupId>
-      <artifactId>tape</artifactId>
-      <version>1.2.3</version>
-    </dependency>
-    <dependency>
-      <groupId>com.github.ben-manes.caffeine</groupId>
-      <artifactId>caffeine</artifactId>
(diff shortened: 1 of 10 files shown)


proxy/pom.xml
```diff
@@ -87,16 +87,6 @@
       <version>3.1</version>
     </dependency>
     <dependency>
-      <groupId>com.squareup</groupId>
```
panghy: I am guessing that it's because it's already inherited?

basilisk487: The "tape" dependency was simply a duplicate in the proxy's pom.xml, and two "caffeine" dependencies were referring to different version, so maven was complaining about both of them.

PR #118, opened 2016-11-03
Limit memory buffers based on heap size

basilisk487:


moribellamy:
@panghy can you take a look at this one?

PR #123, opened 2016-11-08
WavefrontHistogram now supports a bulk-update.

timcschmidt:


java-lib/src/main/java/com/yammer/metrics/core/WavefrontHistogram.java
```diff
@@ -90,17 +91,40 @@ public void update(int value) {
...
+  }
+
+  /**
+   * Bulk-update this histogram with a set of centroids.
+   *
+   * @param means the centroid values
+   * @param counts the centroid weights/sample counts
+   */
+  public synchronized void bulkUpdate(List<Double> means, List<Integer> counts) {
+    if (means != null &&  counts != null) {
```
panghy: just some formatting glitches (also 119)

PR #124, opened 2016-11-09
Don't try to escape double quotes if none present

basilisk487:


panghy:
Why is this change necessary?

basilisk487:
As it turned out, on loads that use point tags, escapeQuotes uses more CPU than pointToStringSB itself (see jvmtop output below) - and it's only useful in fringe cases. So, this was a "low hanging fruit" opportunity to further reduce CPU usage. 

```
  15.53% (    30.63s) org.antlr.v4.runtime.atn.LexerATNSimulator.execATN()
  10.67% (    21.04s) com.wavefront.agent.PostPushDataTimedTask.addPoint()
   8.76% (    17.28s) org.antlr.v4.runtime.BufferedTokenStream.getText()
   7.57% (    14.92s) com.wavefront.ingester.IngesterFormatter.getLiteral()
   6.35% (    12.53s) com.wavefront.agent.PointHandlerImpl.escapeQuotes()
   5.21% (    10.28s) com.wavefront.ingester.IngesterFormatter.parseValue()
   5.20% (    10.25s) com.wavefront.agent.PointHandlerImpl.pointToStringSB()
```

panghy:
I don't think doing something O(N) twice going to be faster than what StringUtils is already doing (it has an indexOf() optimization which is what contains() would have done).

I think jvmtop is misleading in getting accurate traces (that's the woe of doing sampling), try yourkit tracing if we want to look deeper into what's really going on.

basilisk487:
Agreed - the CPU profile does appear misleading, as StringUtils already performs that check. Closing PR.

PR #130, opened 2016-12-06
Misc improvements

basilisk487:
1. InetAddress.getLocalHost() returns 127.0.1.1 on Debian - work around this issue by using the address from the first connected network interface, prefer IPv4 if available.
2. Cap auto-configured number of threads at 16, so instances running in containers on powerful hardware won't spawn too many threads
3. Disable automatic retries in HTTPClient so push threads can regain control quickly

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
+++ b/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
@@ -402,14 +402,17 @@ public abstract class AbstractAgent {
     @Override
     public void run() {
       long startTime = System.currentTimeMillis();
+      boolean isRetry = false;
       try {
         AgentConfiguration config = fetchConfig();
         if (config != null) {
           processConfiguration(config);
+        } else {
+          isRetry = true;
         }
       } finally {
-        // schedule the next run in 1 minute, compensated for the time taken to check in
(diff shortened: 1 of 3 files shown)


proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
```diff
@@ -1065,27 +1069,31 @@ public void shutdown() {
...
-        // if can't resolve local server name, use a real IPv4 address from the first available network interface
-        for (Enumeration<NetworkInterface> nics = NetworkInterface.getNetworkInterfaces();
-             nics.hasMoreElements(); ) {
-          NetworkInterface network = nics.nextElement();
-          if (!network.isUp() || network.isLoopback())
+      Enumeration<NetworkInterface> nics = NetworkInterface.getNetworkInterfaces();
+      while (nics.hasMoreElements()) {
+        NetworkInterface network = nics.nextElement();
+        if (!network.isUp() || network.isLoopback())
+          continue;
```
moribellamy: either use braces or keep on same line, plz

basilisk487: @moribellamy done

PR #135, opened 2016-12-15
Attempt to auto-reconnect on broken pipe error

basilisk487:


java-client/src/main/java/com/wavefront/integrations/Wavefront.java
```diff
@@ -142,6 +143,21 @@ private void internalSend(String name, double value, @Nullable Long timestamp, S
...
       }
     }
+    try {
+      internalSend0(name, value, timestamp, source, pointTags);
+    } catch (SocketException ex) { // retry at least once on SocketExceptions
+      this.socket = null;
+      try {
+        connect();
+      } catch (IllegalStateException e) {
+        // already connected.
```
moribellamy: you should probably log both the SocketException and the IllegalStateException.

Even if you're pretty sure you only get IllegalStateException iff "already connected", probably a good idea to log just to be sure

java-client/src/main/java/com/wavefront/integrations/Wavefront.java
```diff
@@ -142,6 +143,21 @@ private void internalSend(String name, double value, @Nullable Long timestamp, S
...
+      }
+      internalSend0(name, value, timestamp, source, pointTags);
+    }
+  }
+
+  private void internalSend0(String name, double value, @Nullable Long timestamp, String source,
```
moribellamy: better names please. maybe "internalSendWithOneRetry" and "internalSend", or come up with something better. internalSend0 isn't too helpful

java-client/src/main/java/com/wavefront/integrations/Wavefront.java
```diff
@@ -170,11 +186,12 @@ private void internalSend(String name, double value, @Nullable Long timestamp, S
           sb.append(' ');
           sb.append(sanitize(tag.getKey()));
           sb.append('=');
-          sb.append(sanitize(tag.getValue()));
+          sb.append(quote(tag.getValue())); // spaces are allowed in point tag values
         }
       }
       sb.append('\n');
       writer.write(sb.toString());
+      writer.flush();
```
moribellamy: are you sure this doesn't have performance implications under load?

java-client/src/main/java/com/wavefront/integrations/Wavefront.java
```diff
@@ -214,11 +231,15 @@ public synchronized void close() throws IOException {
 
   static String sanitize(String s) {
     final String whitespaceSanitized = WHITESPACE.matcher(s).replaceAll("-");
+    return quote(whitespaceSanitized);
+  }
+
+  private static String quote(String s) {
```
moribellamy: you probably did already but please make sure these are covered in unit tests

PR #138, opened 2017-01-26
Orphaned .pid bugfix + JRE download through http proxy 

basilisk487:


pkg/etc/init.d/wavefront-proxy
```diff
@@ -63,6 +63,33 @@ if [[ -r $proxy_launch_conf ]]; then
 	fi
 fi
 
+if [[ ! -r $JAVA_HOME/bin/java ]]; then
```
moribellamy: Few concerns here:

1) If JAVA_HOME is empty, i wonder if some users would have /bin/java, and if entering the loop in that case would be astonishing to them?
2) If their JAVA_HOME indicates a JRE that is specific to their deployment, but too old for wavefront (e.g. java7), then the customer will not be happy with this setup

What do you think of having something like WAVEFRONT_JAVA_HOME to indicate that the customer has opted into using their own JRE?

basilisk487: JAVA_HOME is actually overridden in the script to /opt/wavefront/wavefront-proxy/proxy-jre [(link)](https://github.com/basilisk487/java/blob/5600e881ee1c999e54c0a88c15745b3db4f52964/pkg/etc/init.d/wavefront-proxy#L29), so I believe these scenarios are not applicable here. 

However, it is indeed a good idea to allow user-defined JREs as well - added as $PROXY_JAVA_HOME.

PR #143, opened 2017-02-01
Fix missing ~agent.buffer.fill-rate metric

basilisk487:
Background: getPostingSizerTask would quietly throw an InvalidObjectState exception on .finish() before it had a chance to update the histogram with sizes


jbau:
LGTM.

PR #144, opened 2017-02-03
Set log files owner to proxy user 

basilisk487:
(PUB-57 bugfix)

### DIFF
--- a/pkg/after-install.sh
+++ b/pkg/after-install.sh
@@ -23,6 +23,18 @@ fi
 # Create log directory if it does not exist
 [[ -d $log_dir ]] || mkdir -p $log_dir && chown $user:$group $log_dir
 
+# if jsvc logs don't exist, create empty files
+if [[ ! -f $log_dir/wavefront-daemon.log ]]; then
+    > $log_dir/wavefront-daemon.log
+fi
+
+if [[ ! -f $log_dir/wavefront-error.log ]]; then
+    > $log_dir/wavefront-error.log
+fi
+
+chown $user:$group $log_dir/wavefront-daemon.log
+chown $user:$group $log_dir/wavefront-error.log
+
 # Configure agent to start on reboot.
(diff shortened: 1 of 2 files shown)


pkg/after-install.sh
```diff
@@ -23,6 +23,18 @@ fi
...
 [[ -d $log_dir ]] || mkdir -p $log_dir && chown $user:$group $log_dir
 
+# if jsvc logs don't exist, create empty files
+if [[ ! -f $log_dir/wavefront-daemon.log ]]; then
+    > $log_dir/wavefront-daemon.log
+fi
+
+if [[ ! -f $log_dir/wavefront-error.log ]]; then
+    > $log_dir/wavefront-error.log
+fi
```
moribellamy: probably fine to just use `touch` without an if-guard, but i guess that would modify the mtime and maybe you don't want that.

pkg/etc/init.d/wavefront-proxy
```diff
@@ -102,7 +102,14 @@ jsvc=$proxy_dir/bin/jsvc
 
 jsvc_exec()
 {
+    if [[ ! $1 == "-stop" ]]; then
+        > $daemon_log_file
```
moribellamy: don't you want to append to this file if it already exists, instead of overwriting it each time?

pkg/after-install.sh
basilisk487: Agreed - `touch` is cleaner.

pkg/etc/init.d/wavefront-proxy
basilisk487: Daemon log has always been overwritten, since it was a redirect of stdout anyway. These two logs have very limited use (we haven't had to use them yet), so I believe that keeping these logs just for the last run is sufficient - one less thing to maintain and rotate.

PR #145, opened 2017-02-09
Interactive grok debugging

moribellamy:
In a new mode, read lines from STDIN and for each line print every match for that line in logsIngestionConfig.yaml

e.g. java -jar target/proxy-4.5-SNAPSHOT-uber.jar -f config.ini --testLogs < foo.txt

where foo.txt has, e.g.
```
notmatched
counter bar 2
counter foo 42
...
```

May also omit the last bit "< foo.txt" to use interactive terminal mode

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/logsharvesting/InteractiveLogsTester.java
+++ b/proxy/src/main/java/com/wavefront/agent/logsharvesting/InteractiveLogsTester.java
@@ -1,0 +1,84 @@
+package com.wavefront.agent.logsharvesting;
+
+import com.wavefront.agent.PointHandler;
+import com.wavefront.agent.config.ConfigurationException;
+import com.wavefront.agent.config.LogsIngestionConfig;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+import java.util.Scanner;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
+
+import javax.annotation.Nullable;
+
+import sunnylabs.report.ReportPoint;
+
+/**
+ * @author Mori Bellamy (mori@wavefront.com)
+ */
+public class InteractiveLogsTester {
+
+  private final Supplier<LogsIngestionConfig> logsIngestionConfigSupplier;
+  private final String prefix;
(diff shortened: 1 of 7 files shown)


proxy/src/main/java/com/wavefront/agent/logsharvesting/InteractiveLogsTester.java
```diff
@@ -0,0 +1,84 @@
...
+   */
+  public boolean interactiveTest() throws ConfigurationException {
+    final AtomicBoolean reported = new AtomicBoolean(false);
+
+    LogsIngester logsIngester = new LogsIngester(
+        new PointHandler() {
+          @Override
+          public void reportPoint(ReportPoint point, @Nullable String debugLine) {
+            reported.set(true);
+            System.out.println(point.toString());
```
basilisk487: Maybe use PointHandlerImpl.pointToString() instead to output data in a format more familiar to the user? That method is begging to be moved elsewhere (or at least be made public).

proxy/src/main/java/com/wavefront/agent/logsharvesting/LogsIngester.java
```diff
@@ -64,13 +64,21 @@ public LogsIngester(PointHandler pointHandler, Supplier<LogsIngestionConfig> log
...
     this.pointHandler = pointHandler;
 
     // Continually flush user metrics to Wavefront.
     this.metricsReporter = new MetricsReporter(
         evictingMetricsRegistry.metricsRegistry(), flushProcessor, "FilebeatMetricsReporter", pointHandler, prefix);
-    this.metricsReporter.start(config.aggregationIntervalSeconds, TimeUnit.SECONDS);
+  }
+
+  public void start() {
+    this.metricsReporter.start(
```
basilisk487: This is obviously not critical, but it would be nice to print a message as it begins reading from stdin to let the user know it's ready to accept data

proxy/src/main/java/com/wavefront/agent/logsharvesting/InteractiveLogsTester.java
moribellamy: done

PR #157, opened 2017-03-13
Prevent buffer file corruption in case of concurrent access (PUB-65)

basilisk487:


proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
```diff
@@ -156,33 +160,44 @@ public QueuedAgentService(AgentAPI service, String bufferFile, final int retryTh
...
+          LZ4BlockOutputStream lz4BlockOutputStream = new LZ4BlockOutputStream(bytes);
+          ObjectOutputStream oos = new ObjectOutputStream(lz4BlockOutputStream);
+          oos.writeObject(o);
+          oos.close();
+          lz4BlockOutputStream.close();
+        }
+      });
+
+      try {
+        FileChannel channel = new RandomAccessFile(buffer, "rw").getChannel();
```
panghy: You also found this trick, please put a comment that this doesn't work on all platforms

proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
```diff
@@ -156,33 +160,44 @@ public QueuedAgentService(AgentAPI service, String bufferFile, final int retryTh
...
+        Preconditions.checkNotNull(channel.tryLock());
+      } catch (Exception e) {
+        logger.severe("Error initializing buffer file " + bufferFile + "." + i +
```
panghy: Add error code and document (see other PR)

PR #170, opened 2017-05-03
Compatibility changes to WavefrontYammerMetricsReporter:

moribellamy:
1) Optionally add JVM metrics, similarly to JsonMetricsGenerator.
   Refactor that bit to use common code.
2) Break timers into their duration/rate parts, similarly to
   JsonMetricsGenerator and DISsimilarly to WavefrontDropwizardMetricsReporter.

### DIFF
--- a/yammer-metrics/src/test/java/com/wavefront/integrations/metrics/WavefrontYammerMetricsReporterTest.java
+++ b/yammer-metrics/src/test/java/com/wavefront/integrations/metrics/WavefrontYammerMetricsReporterTest.java
@@ -31,6 +31,7 @@ import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.startsWith;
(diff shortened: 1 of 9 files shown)


yammer-metrics/src/test/java/com/wavefront/integrations/metrics/WavefrontYammerMetricsReporterTest.java
```diff
@@ -190,27 +206,27 @@ public void testPlainTimer() throws Exception {
...
-        startsWith("\"mytimer.median\""),
-        startsWith("\"mytimer.p75\""),
-        startsWith("\"mytimer.p95\""),
-        startsWith("\"mytimer.p99\""),
-        startsWith("\"mytimer.p999\""),
-        startsWith("\"mytimer.m1\""),
-        startsWith("\"mytimer.m5\""),
-        startsWith("\"mytimer.m15\""),
-        startsWith("\"mytimer.mean\"")
+        startsWith("\"mytimer.duration.min\""),
```
panghy: hm, we may need to convey this change to users or, alternatively, we may need to stick to the old behavior

PR #186, opened 2017-06-15
Changed the proxy behavior so that if the agent_configuration.histogram_disabled, then don't ingest histogram points

sushantdewan123:
- Changed the proxy behavior so that if the agent_configuration.histogram_disabled then don't bother sending the histogram points to the server.
- Reject the points at QueuingChannelHandler.channelRead phase itself. The agent configuration is periodically polled in the checkin API and updated on the proxy.

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
+++ b/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
@@ -582,7 +582,7 @@ public abstract class AbstractAgent {
 
   // Will be updated inside processConfiguration method and the new configuration is periodically
   // loaded from the server by invoking agentAPI.checkin
-  protected AtomicBoolean histogramDisabled = new AtomicBoolean(false);
+  protected final AtomicBoolean histogramDisabled = new AtomicBoolean(false);
 
   /**
    * Executors for support tasks.
--- a/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
+++ b/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
@@ -83,6 +83,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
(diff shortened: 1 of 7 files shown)


proxy/src/main/java/com/wavefront/agent/histogram/QueuingChannelHandler.java
```diff
@@ -48,8 +58,15 @@ private void ship() {
   }
 
   private void innerAdd(T t) {
-    synchronized (this) {
-      buffer.add(t);
+    if (histogramDisabled.get()) {
+      // if histogram feature is disabled on the server increment counter and log it ...
+      logger.info("Ingested point discarded because histogram feature is disabled on the server");
```
jbau: This might be super chatty. is there a way we can only report it every once in a while?

proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -620,6 +621,10 @@ protected void processConfiguration(AgentConfiguration config) {
         retryBackoffBaseSeconds.set(retryBackoffBaseSecondsInitialValue);
         logger.fine("Agent backoff base set to (locally) " + retryBackoffBaseSeconds.get());
       }
+
+      if (config.getHistogramDisabled() != null) {
```
jbau: maybe we want to set the AtomicBoolean to false if the config is null also?

jbau:
i guess the top level question is whether we can just shut down the listener on that histogram port altogether with a flag and then reenable it if the flag is flipped?

proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
```diff
@@ -579,6 +580,10 @@
   protected final boolean localAgent;
   protected final boolean pushAgent;
 
+  // Will be updated inside processConfiguration method and the new configuration is periodically
+  // loaded from the server by invoking agentAPI.checkin
+  protected AtomicBoolean histogramDisabled = new AtomicBoolean(false);
```
panghy: final?

proxy/src/main/java/com/wavefront/agent/histogram/QueuingChannelHandler.java
```diff
@@ -48,8 +61,18 @@ private void ship() {
   }
 
   private void innerAdd(T t) {
-    synchronized (this) {
-      buffer.add(t);
+    if (histogramDisabled.get()) {
+      // if histogram feature is disabled on the server increment counter and log it every 25 times ...
+      discardedHistogramPointsCounter.inc();
+      if (channelReadCounter == 0) {
```
panghy: Just use a ratelimiter and put 0.2 as the rate (once every 5s).

PR #210, opened 2017-09-21
Make WavefrontHistogram more performant by reducing contention on synchronized code

sushantdewan123:
Fix for MONIT-7176

### DIFF
--- a/java-lib/src/main/java/com/yammer/metrics/core/WavefrontHistogram.java
+++ b/java-lib/src/main/java/com/yammer/metrics/core/WavefrontHistogram.java
@@ -13,8 +13,8 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.function.Supplier;
 
 import static com.google.common.collect.Iterables.getFirst;
@@ -33,7 +33,7 @@ public class WavefrontHistogram extends Histogram implements Metric {
   private final static int MAX_BINS = 10;
   private final Supplier<Long> millis;
 
-  private final Map<Long, LinkedList<MinuteBin>> map = new ConcurrentHashMap<>();
+  private final ConcurrentMap<Long, LinkedList<MinuteBin>> perThreadHistogramBins = new ConcurrentHashMap<>();
 
   private WavefrontHistogram(TDigestSample sample, Supplier<Long> millis) {
     super(sample);
@@ -66,10 +66,11 @@ public class WavefrontHistogram extends Histogram implements Metric {
    * @param clear if set to true, will clear the older bins
    * @return returns aggregated collection of all the bins prior to the current minute
    */
-  public Collection<MinuteBin> bins(boolean clear) {
-    Collection<MinuteBin> result = new LinkedList<>();
+  public List<MinuteBin> bins(boolean clear) {
+    List<MinuteBin> result = new LinkedList<>();
     final long cutoffMillis = minMillis();
-    map.values().stream().flatMap(List::stream).filter(i -> i.getMinMillis() < cutoffMillis).forEach(result::add);
+    perThreadHistogramBins.values().stream().flatMap(List::stream).
+        filter(i -> i.getMinMillis() < cutoffMillis).forEach(result::add);
 
     if (clear) {
       clearPriorCurrentMinuteBin(cutoffMillis);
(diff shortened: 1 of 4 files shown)


java-lib/src/main/java/com/yammer/metrics/core/WavefrontHistogram.java
```diff
@@ -91,18 +96,30 @@ public void update(int value) {
...
-   * Helper to retrieve the current bin. Assumes proper synchronization.
+   * Helper to retrieve the current bin. Will be invoked per thread.
    */
   private MinuteBin getCurrent() {
-    long minMillis = minMillis();
+    long currMinMillis = minMillis();
+
+    long key = Thread.currentThread().getId();
+    LinkedList<MinuteBin> bins;
+    if (map.containsKey(key)) {
```
panghy: this can be optimized as just a get() and checking if it's null

java-lib/src/main/java/com/yammer/metrics/core/WavefrontHistogram.java
```diff
@@ -91,18 +96,30 @@ public void update(int value) {
...
+      bins = map.get(key);
+    } else {
+      bins = new LinkedList<>();
```
panghy: this code is problematic, given that this can be called concurrently, you need to use putIfAbsent() and check the result to make sure another thread isn't also doing the same thing

java-lib/src/main/java/com/yammer/metrics/core/WavefrontHistogram.java
```diff
@@ -29,27 +31,13 @@
...
-
-    MinuteBin(long minMillis) {
-      dist = new AVLTreeDigest(ACCURACY);
-      this.minMillis = minMillis;
-    }
 
-    public TDigest getDist() {
-      return dist;
-    }
+  private final Map<Long, LinkedList<MinuteBin>> map = new ConcurrentHashMap<>();
```
panghy: might want to name this something other than just "map"

panghy: Use ConcurrentMap as the type so you can do atomic operations on the map

java-lib/src/main/java/com/yammer/metrics/core/WavefrontHistogram.java
```diff
@@ -61,28 +49,45 @@ public static WavefrontHistogram get(MetricsRegistry registry, MetricName metric
...
-  }
+  /**
+   * Aggregates all the bins prior to the current minute
+   * This is because threads might be updating the current minute bin while the bins() method is invoked
+   *
+   * @param clear if set to true, will clear the older bins
+   * @return returns aggregated collection of all the bins prior to the current minute
+   */
+  public Collection<MinuteBin> bins(boolean clear) {
+    Collection<MinuteBin> result = new LinkedList<>();
```
panghy: Why does this need to be a linked list?

java-lib/src/main/java/com/yammer/metrics/core/WavefrontHistogram.java
```diff
@@ -61,28 +49,45 @@ public static WavefrontHistogram get(MetricsRegistry registry, MetricName metric
...
-    this.bins = new LinkedList<>();
-  }
+  /**
+   * Aggregates all the bins prior to the current minute
+   * This is because threads might be updating the current minute bin while the bins() method is invoked
+   *
+   * @param clear if set to true, will clear the older bins
+   * @return returns aggregated collection of all the bins prior to the current minute
+   */
+  public Collection<MinuteBin> bins(boolean clear) {
```
panghy: It's always preferable to return a more specific collection interface as a return type (while allowing the broadest collection interface as inputs). In your case, I would expect it to return a List<>

java-lib/src/main/java/com/yammer/metrics/core/WavefrontHistogram.java
```diff
@@ -61,28 +49,46 @@ public static WavefrontHistogram get(MetricsRegistry registry, MetricName metric
...
-  }
+  /**
+   * Aggregates all the bins prior to the current minute
+   * This is because threads might be updating the current minute bin while the bins() method is invoked
+   *
+   * @param clear if set to true, will clear the older bins
+   * @return returns aggregated collection of all the bins prior to the current minute
+   */
+  public List<MinuteBin> bins(boolean clear) {
+    List<MinuteBin> result = new LinkedList<>();
```
panghy: can use arraylist

java-lib/src/main/java/com/yammer/metrics/core/WavefrontHistogram.java
```diff
@@ -91,18 +97,26 @@ public void update(int value) {
...
-    long minMillis = minMillis();
-
-    if (bins.isEmpty() || bins.getLast().minMillis != minMillis) {
-      bins.add(new MinuteBin(minMillis));
-      if (bins.size() > MAX_BINS) {
-        bins.removeFirst();
+    long currMinMillis = minMillis();
+
+    long key = Thread.currentThread().getId();
+    LinkedList<MinuteBin> bins = perThreadHistogramBins.getOrDefault(key, new LinkedList<>());
```
panghy: unfortunately you shouldn't do this since every get on an existing key would create a new linkedlist that would be discarded right away

java-lib/src/main/java/com/yammer/metrics/core/WavefrontHistogram.java
```diff
@@ -91,18 +97,26 @@ public void update(int value) {
...
+    perThreadHistogramBins.putIfAbsent(key, bins);
```
panghy: you need to check the output of this since you should be using the return value if it already exists (this doesn't happen since the thread id is unique to the thread I understand)

the typical paradigm for concurrent map is:
```
V value = map.get(key)
if (value == null) {
  // create a new value (only in high contention would this be created but discarded)
  value = new ...
  V existing = map.putIfAbsent(key, value);
  if (existing != null) value = existing;
}
```

PR #214, opened 2017-10-06
Java 9 compatibility 

basilisk487:
Work around the issue in com.yammer.metrics.core.VirtualMachineMetrics: Java 9 is much more restrictive concerning reflective access to non-public fields/methods, as a result the proxy can't generate JVM metrics and register itself. This problem is specific to Mac installations, as we don't use a sterile environment with a bundled JRE there.


panghy:
we should also look into changing to dropwizard metrics (if they now support tags)

PR #225, opened 2017-11-30
Add delta support for counters

sushantdewan123:
Made proxy changes to allow ingesting delta (∆) metrics.


panghy:
Please add tests

PR #230, opened 2017-12-18
Reduce queue locks + non-blocking buffer drains

basilisk487:


proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
```diff
@@ -316,7 +318,7 @@ public Long value() {
 
   private Runnable createRunnable(final ScheduledExecutorService executorService, final boolean
       splitPushWhenRateLimited, final int threadId, final
-      ResubmissionTaskQueue taskQueue, final RecyclableRateLimiter pushRateLimiter) {
+                                  ResubmissionTaskQueue taskQueue, final RecyclableRateLimiter pushRateLimiter) {
```
panghy: alignment is odd

basilisk487:
@panghy Done

PR #231, opened 2017-12-18
Reduce overhead for point logging

basilisk487:


proxy/src/main/java/com/wavefront/agent/PointHandlerImpl.java
```diff
@@ -58,11 +71,17 @@ public PointHandlerImpl(final String handle,
     this.blockedPointsPerBatch = blockedPointsPerBatch;
     this.prefix = prefix;
     String logPointsProperty = System.getProperty("wavefront.proxy.logpoints");
-    this.logPoints = logPointsProperty != null && logPointsProperty.equalsIgnoreCase("true");
+    this.logPointsFlag = logPointsProperty != null && logPointsProperty.equalsIgnoreCase("true");
 
     this.receivedPointLag = Metrics.newHistogram(new MetricName("points." + handle + ".received", "", "lag"));
 
     this.sendDataTasks = sendDataTasks;
+    this.pointHandlerExecutor =  Executors.newScheduledThreadPool(1,
```
panghy: this seems a bit over-complicated (to take a thread to do this). perhaps just add a volatile long and check the loggable status if it's stale instead?

basilisk487:
@panghy Done

proxy/src/main/java/com/wavefront/agent/PointHandlerImpl.java
```diff
@@ -33,15 +33,11 @@
...
-  /**
-   * Executor for support tasks (refreshing logger level)
-   */
-  private final ScheduledExecutorService pointHandlerExecutor;
-
   private final Histogram receivedPointLag;
   private final String validationLevel;
   private final String handle;
   private boolean logPoints = false;
+  private long logPointsUpdatedMillis = 0L;
```
panghy: volatile to be absolutely correct. setting the value of a long might not be atomic

panghy: oh you did it later

PR #233, opened 2017-12-18
Cache queue sizes (reduce I/O on queue files)

basilisk487:


proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
```diff
@@ -162,6 +173,12 @@ public QueuedAgentService(WavefrontAPI service, String bufferFile, final int ret
     String bufferFileSourceTag = bufferFile + "SourceTag";
     this.executorService = executorService;
 
+    queueSizes = Caffeine.newBuilder()
+        .maximumSize(retryThreads)
```
panghy: doesn't seem like this size limit is needed (it's not going to explode and setting a size has a slight performance hit)

basilisk487:
@panghy Done

PR #239, opened 2018-01-11
Yooh/pub 110

howardyoo:
https://wavefront.atlassian.net/browse/PUB-110
This PR adds a configuration to proxy which enabled it to specify pre hour limit which is originally defaulted to 24 hours. With this enhancement, user can now specify a specific hours of which proxy will receive the future timestamps on the point data. Please note however, that the default limit is set to 24 hours (1 day) and user should be careful not to set too large a value, as enabling to ingest future time series can seriously impact the performance.

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/preprocessor/ReportPointTimestampInRangeFilter.java
+++ b/proxy/src/main/java/com/wavefront/agent/preprocessor/ReportPointTimestampInRangeFilter.java
@@ -15,10 +15,9 @@ import wavefront.report.ReportPoint;
 /**
  * Filter condition for valid timestamp - should be no more than 1 day in the future
  * and no more than X hours (usually 8760, or 1 year) in the past
- * 
- * now have additional cutOff values where 
  *
  * Created by Vasily on 9/16/16.
+ * Updated by Howard on 1/10/18 to add support for preCutOffHours
  */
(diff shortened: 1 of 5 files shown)


proxy/src/main/java/com/wavefront/agent/preprocessor/ReportPointTimestampInRangeFilter.java
```diff
@@ -17,16 +17,20 @@
  * and no more than X hours (usually 8760, or 1 year) in the past
  *
  * Created by Vasily on 9/16/16.
+ * Updated by Howard on 1/10/18 to add support for preCutOffHours
  */
 public class ReportPointTimestampInRangeFilter extends AnnotatedPredicate<ReportPoint> {
 
-  private final int cutoffHours;
+  private final int backCutoffHours;
```
panghy: Can we name this better? hoursInThePastToAllow / hoursIntheFutureToAllow

panghy:
The build is broken as well

howardyoo:
Found out a missing parentheses in the filter. Now it should all be good to build. As for the changes in the variable name, Asked @basilisk487 about it, and awaiting his feedback.

howardyoo:
Changed the variable names to "hoursInPastAllowed" and "hoursInFutureAllowed" which made the code much more readable.

PR #288, opened 2018-07-02
Datadog format support + optimize Netty pipeline

basilisk487:


proxy/src/main/java/com/wavefront/agent/CachingGraphiteHostAnnotator.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+@ChannelHandler.Sharable
+public class CachingGraphiteHostAnnotator {
+  private final LoadingCache<InetAddress, String> rdnsCache;
+  private final boolean disableRdnsLookup;
+  private final List<String> sourceTags;
+
+  public CachingGraphiteHostAnnotator(final List<String> customSourceTags, boolean disableRdnsLookup) {
+    this.disableRdnsLookup = disableRdnsLookup;
+    this.sourceTags = Lists.newArrayListWithExpectedSize(customSourceTags.size() + 2);
```
vikramraman: Could customSourceTags be null? If so, add null checks, ignore this comment otherwise.

basilisk487: It's always initialized so it's never null in practice, but you have a very good point - null is a legitimate use case here. Added handling for null cases. 
And, of course, you're right - it should be +4.

PR #296, opened 2018-07-24
PUB-136: improve histogram resilience in gross misconfiguration scenarios

basilisk487:
- Increase default maxBloatSize to 1000
- Make ChronicleMap allocation failure error message more user friendly 
- Additional metrics to detect increased IO caused by memory cache pressure (also due to incorrectly configured accumulator size)

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/histogram/accumulator/AccumulationCache.java
+++ b/proxy/src/main/java/com/wavefront/agent/histogram/accumulator/AccumulationCache.java
@@ -1,6 +1,7 @@
 package com.wavefront.agent.histogram.accumulator;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.RateLimiter;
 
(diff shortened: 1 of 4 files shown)


proxy/src/main/java/com/wavefront/agent/histogram/accumulator/AccumulationCache.java
```diff
@@ -85,21 +94,36 @@ public void delete(@Nonnull HistogramKey key, @Nullable AgentDigest value, @Nonn
...
+                    return digestB;
+                  }
                 } else {
-                  digestB.add(digestA);
-                  return digestB;
+                  return (digestB == null ? digestA : digestB);
+                }
+              });
+            } catch (IllegalStateException e) {
+              if (e.getMessage().contains("Attempt to allocate")) {
```
panghy: This is brittle, can we add a test?

basilisk487: @panghy done

PR #305, opened 2018-08-14
Wavefront Proxy raw data logging

howardyoo:
What the change is about:
This PR will modify ChannelStringHandler.java to have the ability to sample and log RAW data string that is coming into Wavefront Proxy, before the data gets pre-processed. This will allow users to monitor and troubleshoot their incoming data in a raw form that will make it easy to detect any incorrect or missed out pre-processing rules or simply see if their source data is in the right format.

This feature can be enabled by setting the following system properties when starting up the proxy:
wavefront.proxy.lograwdata = (true|false)
wavefront.proxy.lograwdata.sample-rate = (double value, between 0.0d ~ 1.0d)

Also, you can enable the logging behavior dynamically by setting the "RawDataLogger" in java logging to "FINE" level. Log will be suppressed when either the logging level is not FINE, or the wavefront.proxy.lograwdata is set to false.

The logging works by the set sample-rate, which can be adjusted to control how much the raw data would be 'sampled,' as RAW data can be overwhelming when large data are incoming. For example, if you wish to sample only 1% of the overall data, set this to 0.01. If you wish to sample all of the RAW data, you can set it to 1.0. Setting the rate to 0.0 will not produce any logs.

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/listeners/ChannelStringHandler.java
+++ b/proxy/src/main/java/com/wavefront/agent/listeners/ChannelStringHandler.java
@@ -9,14 +9,21 @@ import com.wavefront.agent.preprocessor.ReportableEntityPreprocessor;
 import com.wavefront.ingester.Decoder;
 
 import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.util.List;
+import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nullable;
 
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.SimpleChannelInboundHandler;
+
+import org.apache.commons.lang.math.NumberUtils;
+
 import wavefront.report.ReportPoint;
 
 /**
@@ -29,8 +36,10 @@ import wavefront.report.ReportPoint;
 public class ChannelStringHandler extends SimpleChannelInboundHandler<String> {
 
   private static final Logger blockedPointsLogger = Logger.getLogger("RawBlockedPoints");
+  private static final Logger rawDataLogger = Logger.getLogger("RawDataLogger");
 
   private final Decoder<String> decoder;
+  private static final Random RANDOM = new Random();
 
   /**
    * Transformer to transform each line.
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/com/wavefront/agent/listeners/ChannelStringHandler.java
```diff
@@ -39,17 +48,61 @@
...
+      }
+      logRawUpdatedMillis = System.currentTimeMillis();
+    }
+
+    if ((logRawData || logRawDataFlag) &&
+        (logRawDataRate >= 1.0d || (logRawDataRate > 0.0d && RANDOM.nextDouble() < logRawDataRate))) {
+      if (ctx.channel().remoteAddress() != null) {
+        String hostAddress = ((InetSocketAddress) ctx.channel().remoteAddress()).getAddress().getHostAddress();
+        int localPort = ((InetSocketAddress) ctx.channel().localAddress()).getPort();
+        rawDataLogger.info(String.format("[%s>%d]%s", hostAddress, localPort, msg));
```
basilisk487: Nit-pick: string concatenation is about 10x faster than String.format() so it might make sense to use concatenation here

proxy/src/main/java/com/wavefront/agent/listeners/ChannelStringHandler.java
```diff
@@ -39,17 +48,61 @@
   private final ReportableEntityPreprocessor preprocessor;
   private final PointHandler pointHandler;
 
+  private boolean logRawData = false;
```
panghy: please group fields by type (private static finals together, private finals together then privates)

proxy/src/main/java/com/wavefront/agent/listeners/ChannelStringHandler.java
```diff
@@ -39,17 +48,61 @@
...
+
   public ChannelStringHandler(Decoder<String> decoder,
                               final PointHandler pointhandler,
                               @Nullable final ReportableEntityPreprocessor preprocessor) {
     this.decoder = decoder;
     this.pointHandler = pointhandler;
     this.preprocessor = preprocessor;
+
+    // check the property setting for logging raw data
+    String logRawDataProperty = System.getProperty("wavefront.proxy.lograwdata");
```
panghy: mark @Nullable

proxy/src/main/java/com/wavefront/agent/listeners/ChannelStringHandler.java
```diff
@@ -39,17 +48,61 @@
...
+    logRawDataFlag = logRawDataProperty != null && logRawDataProperty.equalsIgnoreCase("true");
+    String logRawDataSampleRateProperty = System.getProperty("wavefront.proxy.lograwdata.sample-rate");
```
panghy: mark @Nullable

proxy/src/main/java/com/wavefront/agent/listeners/ChannelStringHandler.java
```diff
@@ -39,17 +48,61 @@
...
+    this.logRawDataRate = logRawDataSampleRateProperty != null &&
+        NumberUtils.isNumber(logRawDataSampleRateProperty) ? Double.parseDouble(logRawDataSampleRateProperty) : 1.0d;
+
+    // make sure the rate fits between 0.0d - 1.0d
+    if (logRawDataRate < 0.0d) {
+      logRawDataRate = 0.0d;
```
panghy: should we warn?

proxy/src/main/java/com/wavefront/agent/listeners/ChannelStringHandler.java
howardyoo: done.

proxy/src/main/java/com/wavefront/agent/listeners/ChannelStringHandler.java
howardyoo: good idea!

PR #312, opened 2018-09-10
Support jaeger thrift format ingestion 

basilisk487:


proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -497,6 +507,30 @@ protected void startTraceListener(final String strPort, ReportableEntityHandlerF
...
+            register("Collector::submitBatches", new JaegerThriftCollectorHandler(strPort, handlerFactory,
+                traceDisabled));
+        server.listen().channel().closeFuture().sync();
+        server.shutdown(false);
+      } catch (InterruptedException e) {
+        logger.info("Listener on port " + strPort + " shut down.");
+      } catch (Exception e) {
+        logger.log(Level.SEVERE, "Jaeger trace collector exception", e);
+      } finally {
+        activeListeners.dec();
```
sushantdewan123: @basilisk487 shouldn't activeListeners be a gauge? not part of this PR but you can think about changing this to a gauge in subsequent PR.

Issue #315, opened 2018-09-15
WavefrontReporter supports additional point tags for specific metrics


curtis628:
Currently, the `WavefrontReporter` implementation allows me to include pointTags for all metrics that get exported to my Wavefront instance.

I'd like to add _additional_ point tags to specific metrics based off their name.

I'm interested in tracking response codes for a certain Health API. I might have the following `Meter` names in my `MetricRegistry` tracking the counts/rates of the various return codes for this health API:

* `health_api[status=200]`
* `health_api[status=500]`
* `health_api[status=503]`

Right now, the `WavefrontReporter` "sanitizes" a metric name like this. I'd like to see it be able to parse out key/value pairs within `[%s=%s]` and add them as additional point tags.

If they were point tags, I could create a Query for `ts(health_api)` and see a breakdown of all the different response codes. It would also roll up a "total" API calls metric automatically without me having to track it separately.

Is there a better way to accomplish this? Or is there a reason why this approach would be a bad idea? I'm admittedly no expert when it comes to Wavefront or Dropwizards.

sushantdewan123:
Dropwizard5 RC2 already supports point tags but is not GA yet
https://github.com/dropwizard/metrics/releases/tag/v5.0.0-rc2

If you want you can use this reporter to accomplish this - 
https://github.com/wavefrontHQ/java/tree/master/dropwizard-metrics/dropwizard-metrics5

curtis628:
@sushantdewan123 (or anyone from dropwizard really): I created a pull request that implements what I was looking for: https://github.com/wavefrontHQ/java/pull/318

I ended up doing my own work-around in my own code base which involved:
- copy/pasting existing `WavefrontReporter` implementation in my code base (had to do this due to `WavefrontReporter` having many fields `private`)
- made small changes that you can see if my pull request diff to enable per-metrics reporting

I hate copy/pasting... and figured if I needed it, perhaps someone else might? This is why I drafted the pull request. Please let me know what you think (and feel free to reject it if you don't think it will help other users :)

curtis628:
> Dropwizard5 RC2 already supports point tags but is not GA yet
> https://github.com/dropwizard/metrics/releases/tag/v5.0.0-rc2
> 
> If you want you can use this reporter to accomplish this -
> https://github.com/wavefrontHQ/java/tree/master/dropwizard-metrics/dropwizard-metrics5

[...]

PR #318, opened 2018-09-24
Add reporter that supports point tags on metric-level (#315)

curtis628:
WavefrontPointTagsPerMetricReporter is an extension of the
WavefrontReporter. It offers the same abilities, _plus_ the ability to
associate point tags on a per-metric level based on the metric name.

### DIFF
--- a/dropwizard-metrics/dropwizard-metrics/src/main/java/com/wavefront/integrations/metrics/WavefrontPointTagsPerMetricReporter.java
+++ b/dropwizard-metrics/dropwizard-metrics/src/main/java/com/wavefront/integrations/metrics/WavefrontPointTagsPerMetricReporter.java
@@ -1,0 +1,80 @@
+package com.wavefront.integrations.metrics;
+
+import java.io.IOException;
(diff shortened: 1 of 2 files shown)


curtis628:
This pull request solved the issue I raised: https://github.com/wavefrontHQ/java/issues/315

Vaishnavi109:
Overall looks good to me.

panghy:
@basilisk487 I believe this functionality is in the proxy today, just wanting to confirm?

basilisk487:
@panghy Yes, this can be done at the proxy as well - two example rules are provided in the default preprocessor config: https://github.com/wavefrontHQ/java/blob/master/pkg/etc/wavefront/wavefront-proxy/preprocessor_rules.yaml.default#L71-L75

panghy:
@curtis628 thanks for the PR but at this time, either we are going to ask users to use a metrics library that supports tags or the proxy can be configured to extract them. Obviously nothing prevents you have forking our repo and having this functionality.

PR #328, opened 2018-11-15
populate mandatory service and application span tags

sushantdewan123:


panghy:
@sushantdewan123 why don't we call it jaegar for the application (just so that we don't have to explain what is defaultApplication)

PR #323, opened 2018-10-08
PUB-140: Implement authentication mechanism for incoming HTTP requests

basilisk487:

PR #331, opened 2018-11-21
PUB-145: fix preprocessorConfigfile args issue

sliu2013:
* This is to partially revert this change (https://github.com/wavefrontHQ/java/commit/28f7b924c33f1cc8dda518ce0a565c46b43b1bc4). 

* Because of this line(https://github.com/wavefrontHQ/java/blob/master/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java#L816), if there is no config file ("-f") is provided, then lots of config parameters won't be taken effect. This pull request is just a quick workaround, but we might need to modify AbstractAgent.java file more.

* Or maybe we can simply remove this line (https://github.com/wavefrontHQ/java/blob/master/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java#L816) ?!


sliu2013:
It's included in https://github.com/wavefrontHQ/java/pull/323

PR #339, opened 2018-12-13
MONIT-12118 support zipkin spans

akodali18:
Initial commit to support zipkin traces in wavefront proxy.


proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -580,6 +579,14 @@ protected void startTraceJaegerListener(String strPort, ReportableEntityHandlerF
     logger.info("listening on port: " + strPort + " for trace data (Jaeger format)");
   }
 
+  protected void startTraceZipkinListener(String strPort, ReportableEntityHandlerFactory handlerFactory) {
+    final int port = Integer.parseInt(strPort);
+    ChannelHandler channelHandler = new ZipkinPortUnificationHandler(strPort, handlerFactory, traceDisabled);
+
+    startAsManagedThread(new TcpIngester(createInitializer(channelHandler, strPort), port)
+            .withChildChannelOptions(childChannelOptions), "listener-zipkin-trace-" + port);
```
sushantdewan123: period on line 586 and line break after .

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,260 @@
...
+  private final AtomicBoolean traceDisabled;
+  private final RateLimiter warningLoggerRateLimiter = RateLimiter.create(0.2);
+  private final Counter discardedTraces;
+  private final Counter discardedBatches;
+  private final Counter processedBatches;
+  private final Counter failedBatches;
+
+  private final static Set<String> ZIPKIN_VALID_PATHS = ImmutableSet.of("/api/v1/spans/", "/api/v2/spans/");
+  private final static String ZIPKIN_VALID_HTTP_METHOD = "POST";
+  private final static String APPLICATION_KEY = "application";
```
sushantdewan123: Can you share this across Jaeger and Zipkin? Ideally we should get this from tier 1 SDK
since are adding that dependency already, please use
https://github.com/wavefrontHQ/wavefront-sdk-java/blob/master/src/main/java/com/wavefront/sdk/common/Constants.java#L61

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,260 @@
...
+
+  public ZipkinPortUnificationHandler(final String handle,
+                                      ReportableEntityHandler<Span> handler,
+                                      AtomicBoolean traceDisabled) {
+    super(TokenAuthenticatorBuilder.create().setTokenValidationMethod(TokenValidationMethod.NONE).build(), handle,
+            false, true);
+    this.handle = handle;
+    this.handler = handler;
+    this.traceDisabled = traceDisabled;
+    this.discardedTraces = Metrics.newCounter(new MetricName("spans." + handle, "", "discarded"));
```
sushantdewan123: should be called "discardedSpans" and not "discardedTraces"

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,260 @@
...
+
+    // Validate Uri Path and HTTP method of incoming Zipkin spans.
+    if (!ZIPKIN_VALID_PATHS.contains(path)) {
+      writeHttpResponse(ctx, HttpResponseStatus.BAD_REQUEST, "Unsupported URL path.", incomingRequest);
+      logWarning("WF-400: Requested URI path '" + path + "' is not supported.", null, ctx);
+      return;
+    }
+    if (!incomingRequest.method().toString().equalsIgnoreCase(ZIPKIN_VALID_HTTP_METHOD)) {
+      writeHttpResponse(ctx, HttpResponseStatus.BAD_REQUEST, "Unsupported Http method.", incomingRequest);
+      logWarning("WF-400: Requested http method '" + incomingRequest.method().toString() + "' is not supported.", null, ctx);
```
sushantdewan123: if you use wavefront codeStylesettings.xml, it will line break for you.

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,260 @@
...
+      return;
+    }
+
+    HttpResponseStatus status;
```
sushantdewan123: I suggest you do this check - if (traceDisabled.get()) {
here instead of wasting CPU.
the traceDisabled flag is changed dynamically using feedback from wavefront server checkin method.

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,260 @@
...
+            .setSpanId(getSpanUuid(zipkinSpan))
+            .setTraceId(convertToUuidString(zipkinSpan.traceId()))
+            .setStartMillis(zipkinSpan.timestampAsLong() / 1000)
+            .setDuration(zipkinSpan.durationAsLong() / 1000)
+            .setAnnotations(annotations)
+            .build();
+    handler.report(newSpan);
+  }
+
+    private List<Annotation> addAnnotations(zipkin2.Span zipkinSpan) {
```
sushantdewan123: indentation not correct.

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,260 @@
...
+    for (zipkin2.Span zipkinSpan : zipkinSpans) {
+      processZipkinSpan(zipkinSpan);
+    }
+  }
+
+  private void processZipkinSpan(zipkin2.Span zipkinSpan) {
+    List<Annotation> annotations = addAnnotations(zipkinSpan);
+
+    // TODO: Confirm source value
+    String sourceName = zipkinSpan.tags().get(SOURCE_KEY) == null ? "unknown" : zipkinSpan.tags().get(SOURCE_KEY);
```
sushantdewan123: let's default to "zipkin" instead of "unknown".

sushantdewan123: if the source is provided by the tag, then use that.
else default to localEndpoint.ipv4
and if that is null, default to "Zipkin"

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,260 @@
...
+    if (zipkinSpan.tags().get(CLUSTER_KEY) != null) {
+      annotations.add(new Annotation(CLUSTER_KEY, zipkinSpan.tags().get(CLUSTER_KEY)));
+    }
+    if (zipkinSpan.tags().get(SHARD_KEY) != null) {
+      annotations.add(new Annotation(SHARD_KEY, zipkinSpan.tags().get(SHARD_KEY)));
+    }
+
+    // Set Span's References.
+    if (zipkinSpan.parentId() != null) {
+      annotations.add(new Annotation("parent", convertToUuidString(zipkinSpan.parentId())));
```
sushantdewan123: Can you put "parent" and "followsFrom" in a constants file so that Jaeger and Zipkin can use those constants?

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,260 @@
...
+
+    //Build wavefront span
+    Span newSpan = Span.newBuilder()
+            .setCustomer("dummy")
+            .setName(zipkinSpan.name())
+            .setSource(sourceName)
+            .setSpanId(getSpanUuid(zipkinSpan))
+            .setTraceId(convertToUuidString(zipkinSpan.traceId()))
+            .setStartMillis(zipkinSpan.timestampAsLong() / 1000)
+            .setDuration(zipkinSpan.durationAsLong() / 1000)
```
sushantdewan123: I thought you mentioned duration is in nanos?
if yes, then you need to convert it into millis not micros

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
akodali18: The reason we do this further down is to be able to report "discarded no. of spans" metric values.

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
akodali18: Mentioned incorrectly in the doc. Code does the right thing. (both timestamp and duration are reported by zipkin in micros)

proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -28,18 +28,10 @@
...
 import com.wavefront.agent.histogram.tape.TapeStringListConverter;
-import com.wavefront.agent.listeners.DataDogPortUnificationHandler;
-import com.wavefront.agent.listeners.JaegerThriftCollectorHandler;
-import com.wavefront.agent.listeners.JsonMetricsEndpoint;
-import com.wavefront.agent.listeners.OpenTSDBPortUnificationHandler;
-import com.wavefront.agent.listeners.RelayPortUnificationHandler;
-import com.wavefront.agent.listeners.TracePortUnificationHandler;
-import com.wavefront.agent.listeners.WavefrontPortUnificationHandler;
-import com.wavefront.agent.listeners.WriteHttpJsonMetricsEndpoint;
+import com.wavefront.agent.listeners.*;
```
basilisk487: Pls use correct IntelliJ code style - we are not supposed to collapse imports into wildcards

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,268 @@
...
+    int offset = desiredLength - id.length();
+
+    for (int i = 0; i < offset; i++) builder.append('0');
+    builder.append(id);
+    return builder.toString();
+  }
+
+  private static String addHyphens(String id) {
+    return id.replaceAll(
+        "(\\w{8})(\\w{4})(\\w{4})(\\w{4})(\\w{12})",
```
basilisk487: Precompile regex

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,268 @@
...
+        "$1-$2-$3-$4-$5");
+  }
+
+  private static String convertToUuidString(String id) {
```
basilisk487: Maybe move to Util class?

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,268 @@
...
+    if (mostSig.startsWith("-")) {
+      mostSig = mostSig.substring(1);
+    }
+    spanUuid = UUID.fromString(mostSig.replaceAll("(\\w{8})(\\w{4})(\\w{4})", "$1-$2-$3") +
+        "-" +
+        spanId.replaceAll("(\\w{4})(\\w{12})", "$1-$2")).toString();
+    return spanUuid;
+  }
+
+  private static String padLeft(String id, int desiredLength) {
```
basilisk487: StringUtils.leftPad?

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
akodali18: ok, Removed tracking metrics for no. of discarded spans.

proxy/src/main/java/com/wavefront/agent/PushAgent.java
akodali18: Agree, this was unintended, i have intellij optimize imports turned on which caused this.

java-lib/src/main/java/com/wavefront/common/TraceConstants.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Trace constants.
+ *
+ * @author Anil Kodali (akodali@vmware.com)
+ */
+public abstract class TraceConstants {
+  // Span References
+  // See https://opentracing.io/specification/ for more information about span references.
+  public static final String FOLLOWS_FROM_KEY = "followsFrom";
```
vikramraman: Okay for now. But ideally this should go in `wavefront-sdk-java`.

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,262 @@
...
+  private final RateLimiter warningLoggerRateLimiter = RateLimiter.create(0.2);
+  private final Counter discardedBatches;
+  private final Counter processedBatches;
+  private final Counter failedBatches;
+
+  private final static Set<String> ZIPKIN_VALID_PATHS = ImmutableSet.of(
+      "/api/v1/spans/",
+      "/api/v2/spans/");
+  private final static String ZIPKIN_VALID_HTTP_METHOD = "POST";
+  private final static String APPLICATION_KEY = Constants.APPLICATION_TAG_KEY;
```
vikramraman: You can use static imports instead of redeclaring these constants.

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,262 @@
...
+
+    // Set Span's References.
+    if (zipkinSpan.parentId() != null) {
+      annotations.add(new Annotation(TraceConstants.PARENT_KEY, Utils.convertToUuidString(zipkinSpan
+          .parentId())));
+    }
+
+    // Set Span Http Tags.
+    if (zipkinSpan.tags() != null && zipkinSpan.tags().size() > 0) {
+      annotations.add(new Annotation("span.kind", zipkinSpan.kind().toString().toLowerCase()));
```
vikramraman: It appears `kind()` can return null: https://github.com/openzipkin/zipkin/blob/0efe65a8685c01d2b09da20c94667fb2c07d21c4/zipkin/src/main/java/zipkin2/Span.java#L131

Please handle accordingly to prevent any NPEs.

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,262 @@
...
+      // TODO: Check if these tags are required for a span with span.kind = client.
+      if (!zipkinSpan.kind().toString().equalsIgnoreCase("client")) {
+        annotations.add(new Annotation("http.method", zipkinSpan.tags().get("http.method")));
```
vikramraman: Better to verify these keys exist in the zipkin span tags prior to adding. Maybe a simple helper method that adds the annotation only if the value is non-empty.

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
vikramraman: Also should this be outside the if block? Not sure if this has any ties to tags.

proxy/src/test/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandlerTest.java
```diff
@@ -0,0 +1,143 @@
...
+import static org.easymock.EasyMock.verify;
+
+public class ZipkinPortUnificationHandlerTest {
+  private ReportableEntityHandler<Span> mockTraceHandler =
+      MockReportableEntityHandlerFactory.getMockTraceHandler();
+  private long startTime = System.currentTimeMillis();
+
+  @Test
+  public void testZipkinHandler() throws Exception {
+    ZipkinPortUnificationHandler handler = new ZipkinPortUnificationHandler("9411",
```
basilisk487: Ok for now but let's fix this in the future - this port could be already used, causing tests to fail.

proxy/src/main/java/com/wavefront/agent/Utils.java
```diff
@@ -31,4 +37,32 @@ public T get() {
...
+  public static String addHyphensToUuid(String uuid) {
+    Matcher matcherUuid = patternUuid.matcher(uuid);
+    return matcherUuid.replaceAll("$1-$2-$3-$4-$5");
+  }
+
+  /**
+   * Method converts a string Id to {@code UUID}.
+   * This Method specifically converts id's with less than 32 digit hex characters into UUID
+   * format (See <a href="http://www.ietf.org/rfc/rfc4122.txt"> <i>RFC&nbsp;4122: A
+   * Universally Unique IDentifier (UUID) URN Namespace</i></a>) by left padding id with Zeroes
```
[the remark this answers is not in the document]
akodali18: Picked this up from javadoc in https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html "see RFC 4122: A Universally Unique IDentifier (UUID) URN Namespace" 
Guess the intention is to highlight the letters based on which acronym is formed.

java-lib/src/main/java/com/wavefront/common/TraceConstants.java
sushantdewan123: Can add a TODO for this so that we can track this.

proxy/src/main/java/com/wavefront/agent/Utils.java
```diff
@@ -31,4 +37,34 @@ public T get() {
...
+   * and adding hyphens. It assumes that if the input id contains hyphens it is already an UUID.
+   * Please don't use this method to validate/guarantee your id as an UUID.
+   *
+   * @param id a string encoded in hex characters.
+   * @return a UUID string.
+   */
+  public static String convertToUuidString(String id) {
```
sushantdewan123: if input id is Nullable (i.e. line 65), then mark it as @Nullable
similarly if the method might return null, then mark it as @Nullable

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,276 @@
...
+  private final Counter processedBatches;
+  private final Counter failedBatches;
+
+  private final static Set<String> ZIPKIN_VALID_PATHS = ImmutableSet.of(
+      "/api/v1/spans/",
+      "/api/v2/spans/");
+  private final static String ZIPKIN_VALID_HTTP_METHOD = "POST";
+  private final static String DEFAULT_APPLICATION = "Zipkin";
+  private final static String DEFAULT_SOURCE = "zipkin";
+  private final static String DEFAULT_SERVICE = "unknown";
```
sushantdewan123: Let's call it "defaultService" instead of "unknown"

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,276 @@
...
+  private final static String DEFAULT_SPAN_NAME = "unknown";
```
sushantdewan123: Let's call it "defaultOperation" instead of unknown.

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,276 @@
...
+            "server");
+      }
+      discardedBatches.inc();
+      output.append("Ingested spans discarded because tracing feature is not enabled on the " +
+          "server.");
+      status = HttpResponseStatus.ACCEPTED;
+      writeHttpResponse(ctx, status, output, incomingRequest);
+      return;
+    }
+
```
sushantdewan123: Nit: one line break good enough instead of 2

proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -0,0 +1,276 @@
...
+    List<Annotation> annotations = new ArrayList<>();
+    // Set Span's Application Tags.
+    addTagWithKey(zipkinSpan, annotations, APPLICATION_TAG_KEY, DEFAULT_APPLICATION);
+    addTagWithKey(zipkinSpan, annotations, CLUSTER_TAG_KEY, null);
+    addTagWithKey(zipkinSpan, annotations, SHARD_TAG_KEY, null);
+
+    // Set Span's References.
+    if (zipkinSpan.parentId() != null) {
+      annotations.add(new Annotation(TraceConstants.PARENT_KEY, Utils.convertToUuidString(zipkinSpan
+          .parentId())));
```
sushantdewan123: line break after . instead of before

proxy/src/test/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandlerTest.java
```diff
@@ -0,0 +1,143 @@
...
+            new Annotation("span.kind", "server"),
+            new Annotation("service", "frontend"),
+            new Annotation("http.method", "GET"),
+            new Annotation("http.url", "none+h1c://localhost:8881/"),
+            new Annotation("http.status_code", "200")))
+        .build());
+    expectLastCall();
+
+    mockTraceHandler.report(Span.newBuilder().setCustomer("dummy").setStartMillis(startTime)
+        .setDuration(2234)
```
sushantdewan123: all the periods can occur before the line break
guess we dont follow this consistently else where in our code :(

PR #342, opened 2018-12-14
Added Jaeger span and Wavefront span logging

sushantdewan123:


pkg/etc/wavefront/wavefront-proxy/log4j2.xml.default
```diff
@@ -88,6 +88,12 @@
             <AppenderRef ref="ValidPointsFile"/>
             -->
         </AsyncLogger>
+        <!-- Uncomment AppenderRef and set level="ALL" to log Jaeger data to a file -->
+        <AsyncLogger name="JaegerDataLogger" level="DEBUG" additivity="false">
+            <!--
+            <AppenderRef ref="JaegerDataFile"/>
```
basilisk487: If we're adding a reference to JaegerDataFile, should we define it as well (and comment out)?

PR #347, opened 2019-01-03
Initial commit

sushantdewan123:


proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -196,6 +193,14 @@ protected void startListeners() {
           Collections.emptyIterator() :
           Splitter.on(",").omitEmptyStrings().trimResults().split(histogramDistListenerPorts).iterator();
 
+      // Use a tmp iterator so that we don't modify `histDistPorts` iterator
```
basilisk487: This block is now unnecessary with InternalProxyWavefrontClient; handlerFactory will dynamically allocate a histogram sender and point sender

PR #348, opened 2019-01-07
Handle zipkin errors

akodali18:
Handling zipkin errors. Zipkin spans usually propagate errors with key='error' and value='actual error message'. We want to ignore this error message for now. Instead report it as 'error'='true' so that the span shows up as RED in the UI.


proxy/src/main/java/com/wavefront/agent/listeners/ZipkinPortUnificationHandler.java
```diff
@@ -174,18 +175,28 @@ private void processZipkinSpan(zipkin2.Span zipkinSpan) {
...
+        setTraceId(traceId).
         setStartMillis(zipkinSpan.timestampAsLong() / 1000).
         setDuration(zipkinSpan.durationAsLong() / 1000).
         setAnnotations(annotations).
         build();
 
+    if (zipkinSpan.tags().containsKey(SPAN_TAG_ERROR)) {
+      if (zipkinDataLogger.isLoggable(Level.FINER)) {
+        zipkinDataLogger.info("Span id :: " + spanId + " with trace id :: " + traceId +
+            " , failed with Error :: " + zipkinSpan.tags().get(SPAN_TAG_ERROR));
```
vikramraman: Can we tweak this log message. Change `failed with Error` -> `includes error tag`.

This message makes it seem there is a failure within our proxy.

PR #368, opened 2019-02-27
MONIT-12724  Fix for Proxy with traceSamplingRate unset and traceSamplingDuration effective will generate unexpected result

akodali18:
yet to test.

### DIFF
--- a/pkg/etc/wavefront/wavefront-proxy/wavefront.conf.default
+++ b/pkg/etc/wavefront/wavefront-proxy/wavefront.conf.default
@@ -155,7 +155,6 @@ configuration in Istio.
 
 ## The following settings are used to configure trace data sampling:
 ## The rate for traces to be sampled. Can be from 0.0 to 1.0. Defaults to 1.0
-## Ex: traceSamplingRate = .1 sets up a sampling rate of 10%. A sampling rate of 10% causes 1 out of 10 traces to be sent and ingested.
 #traceSamplingRate=1.0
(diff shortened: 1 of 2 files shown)


pkg/etc/wavefront/wavefront-proxy/wavefront.conf.default
```diff
@@ -153,6 +153,14 @@ buffer=/var/spool/wavefront-proxy/buffer
 configuration in Istio.
 #traceZipkinListenerPorts=9411
 
+## The following settings are used to configure trace data sampling:
+## The rate for traces to be sampled. Can be from 0.0 to 1.0. Defaults to 1.0
+## Ex: The below property sets up a sampling rate of 10%. A sampling rate of 10% causes 1 out of 10 traces to be sent and ingested.
+#traceSamplingRate=.1
```
vikramraman: Let's set them to the default values even when commented out: `1.0` for rate and `0` for duration. (can remove the example part).

pkg/etc/wavefront/wavefront-proxy/wavefront.conf.default
```diff
@@ -153,6 +153,14 @@ buffer=/var/spool/wavefront-proxy/buffer
...
+## The duration for traces to be sampled. Defaults to 0.
```
vikramraman: `duration in milliseconds for the spans to be sampled`.

PR #372, opened 2019-03-28
PUB-151: Don't start backlog processing until first check-in

basilisk487:


proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
```diff
@@ -148,9 +149,9 @@
   @Parameter(names = {"--testLogs"}, description = "Run interactive session for crafting logsIngestionConfig.yaml")
   private boolean testLogs = false;
 
-  @Parameter(names = {"-l", "--loglevel", "--pushLogLevel"}, description =
-      "Log level for push data (NONE/SUMMARY/DETAILED); NONE is default")
-  protected String pushLogLevel = "NONE";
+  @Parameter(names = {"-l", "--loglevel", "--pushLogLevel"}, hidden = true, description =
+      "(DEPRECATED) Log level for push data (NONE/SUMMARY/DETAILED); SUMMARY is default")
```
vikramraman: With this being deprecated, is there a way to get detailed logging? Would we rely on the log4j config for that?

basilisk487: Yes, log4j2 config is the right place to configure logging details. logLevel was a pretty weird way to do detailed logging, tbh

PR #377, opened 2019-05-08
Make AbstractReportableEntityHandler public for SNS Agent usage

song-hao:


song-hao:
Refactored SnsAgent to avoid these changes

PR #378, opened 2019-05-10
Move raw logs ingestion to unified port handler

basilisk487:


pom.xml
```diff
@@ -56,7 +56,7 @@
     <log4j.version>2.11.1</log4j.version>
     <jetty.version>9.4.14.v20181114</jetty.version>
     <jackson.version>2.9.6</jackson.version>
-    <netty.version>4.1.25.Final</netty.version>
+    <netty.version>4.1.36.Final</netty.version>
```
kentwang929: do we want to bump this version higher than what we have in system?

basilisk487: @kentwang929 We usually keep up with netty builds to stay on top of all bugfixes. I would say yes, we want to stay on latest release.

PR #381, opened 2019-05-15
Update default receive buffer size 4KB -> 32KB, update preprocessor r…

basilisk487:
…ule checked counter metric name.


proxy/src/main/java/com/wavefront/agent/preprocessor/AgentPreprocessorConfiguration.java
```diff
@@ -80,7 +80,7 @@ public void loadFromStream(InputStream stream) {
             PreprocessorRuleMetrics ruleMetrics = new PreprocessorRuleMetrics(
                 Metrics.newCounter(new TaggedMetricName("preprocessor." + ruleName, "count", "port", strPort)),
                 Metrics.newCounter(new TaggedMetricName("preprocessor." + ruleName, "cpu_nanos", "port", strPort)),
-                Metrics.newCounter(new TaggedMetricName("preprocessor." + ruleName, "checked.count", "port", strPort)));
+                Metrics.newCounter(new TaggedMetricName("preprocessor." + ruleName, "checked-count", "port", strPort)));
```
basilisk487: This change is needed to avoid confusion when users query `~proxy.preprocessor.*.count`

PR #397, opened 2019-06-25
Added filtering of empty tags in Jaeger and Zipkin spans

asutoshpalai:
Having empty tags is not an error in Jaeger's and Zipkin's spans but those spans get silently rejected by Wavefront server.

This PR filters the empty tags for both the protocol.


proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerThriftCollectorHandler.java
```diff
@@ -241,7 +241,7 @@ private void processSpan(io.jaegertracing.thriftjava.Span span,
 
     if (span.getTags() != null) {
       for (Tag tag : span.getTags()) {
-        if (IGNORE_TAGS.contains(tag.getKey())) {
+        if (IGNORE_TAGS.contains(tag.getKey()) || (tag.vType == TagType.STRING && StringUtils.isBlank(tag.vStr))) {
```
akodali18: Can we use tag.getVStr() instead ?

PR #406, opened 2019-07-13
MONIT-15154: Phase 2 - switch to new backend API (do not merge)

basilisk487:


java-lib/src/main/java/com/wavefront/api/SourceTagAPI.java
```diff
@@ -0,0 +1,94 @@
+package com.wavefront.api;
```
jbau: I assume this file is basically a copy-paste

proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
```diff
@@ -801,8 +768,32 @@ public Response setDescription(String id, String desc, boolean forceToQueue) {
   }
 
   @Override
-  public Response removeTag(String id, String tagValue, boolean forceToQueue) {
+  public Response appendTag(String id, String tagValue, boolean forceToQueue) {
```
jbau: i guess we picked up appendTag as a new API method

PR #407, opened 2019-07-13
MONIT-15154: Phase 1 - new API interface

basilisk487:


proxy/src/main/java/com/wavefront/agent/api/ForceQueueEnabledProxyAPI.java
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Report batched data (metrics, histograms, spans, etc) to Wavefront servers.
+   *
+   * @param proxyId       Proxy Id reporting the result.
+   * @param format        The format of the data (wavefront, histogram, trace, spanLogs)
+   * @param pushData      Push data batch (newline-delimited)
+   * @param forceToQueue  Whether to bypass posting data to the API and write to queue instead.
+   */
+  Response proxyReport(@HeaderParam("X-WF-PROXY-ID") final UUID proxyId,
```
jbau: just a question: this is an API that is on the proxy _input_ side, yes?

basilisk487: @jbau: that's the internal wrapper "API" over proxy's output API that just adds another argument to support writing directly to the on-disk queue instead of sending. 
The output API is in java-lib/src/main/java/com/wavefront/api/ProxyV2API.java - I believe this is a perfect moment to separate these methods into a separate interface so we can eventually deprecate the old ones; and while we're at it, get rid of legacy arguments that no longer make sense.

PR #415, opened 2019-07-19
issue-411 Count spans discarded by sampler

akodali18:
unit tests coming up.

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerThriftCollectorHandler.java
+++ b/proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerThriftCollectorHandler.java
@@ -385,7 +385,7 @@ public class JaegerThriftCollectorHandler extends ThriftRequestHandler<Collector
   }
 
   private boolean sample(Span wavefrontSpan) {
-    if(sampler.sample(wavefrontSpan.getName(),
+    if (sampler != null && sampler.sample(wavefrontSpan.getName(),
         UUID.fromString(wavefrontSpan.getTraceId()).getLeastSignificantBits(),
         wavefrontSpan.getDuration())) {
       return true;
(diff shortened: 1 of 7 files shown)


proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerThriftCollectorHandler.java
```diff
@@ -382,6 +384,16 @@ private void processSpan(io.jaegertracing.thriftjava.Span span,
     }
   }
 
+  private boolean sample(Span wavefrontSpan) {
+    if (sampler != null && sampler.sample(wavefrontSpan.getName(),
```
sushantdewan123: no need to check for sampler is null since it is not marked Nullable
also if at all sampler was null then default is true and not false

proxy/src/main/java/com/wavefront/agent/listeners/tracing/TracePortUnificationHandler.java
```diff
@@ -204,4 +206,13 @@ private static String parseError(String message, @Nullable ChannelHandlerContext
     }
     return errMsg.toString();
   }
+
+  private boolean sample(Span object) {
+    if (sampler != null && sampler.sample(object.getName(),
```
sushantdewan123: if not marked Nullable, dont do null check
also if you do null check then default to true and not false

proxy/src/main/java/com/wavefront/agent/listeners/tracing/ZipkinPortUnificationHandler.java
```diff
@@ -396,6 +399,16 @@ private void processZipkinSpan(zipkin2.Span zipkinSpan) {
     }
   }
 
+  private boolean sample(Span wavefrontSpan) {
+    if (sampler != null && sampler.sample(wavefrontSpan.getName(),
```
sushantdewan123: if not marked Nullable, dont do null check
also if you do null check then default to true and not false

PR #417, opened 2019-07-23
Support heartbeat metric tags for custom derived red metrics

sushantdewan123:


proxy/src/main/java/com/wavefront/agent/listeners/tracing/HeartbeatMetricKey.java
```diff
@@ -15,17 +17,20 @@
...
+  private final String shard;
+  @Nonnull
+  private final String source;
   @Nonnull
-  private String source;
+  private final Map<String, String> customTags;
 
   public HeartbeatMetricKey(String application, String service, String cluster, String shard,
-                            String source) {
+                            String source, Map<String, String> customTags) {
```
basilisk487: If we have these properties annotated as @Nonnull, we should do the same for the constructor arguments

PR #421, opened 2019-07-25
Support debug and sampling.priority span tags for jaeger integration.

akodali18:


proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerThriftCollectorHandler.java
```diff
@@ -245,6 +249,8 @@ private void processSpan(io.jaegertracing.thriftjava.Span span,
     String shard = NULL_TAG_VAL;
     String componentTagValue = NULL_TAG_VAL;
     boolean isError = false;
+    boolean isDebugSpanTag = false;
```
vikramraman: Why not a single boolean to determine whether we should allow the span or not?

akodali18: We just seem to have a lot of factors that affect whether a span should be sampled or not. 
So having separate flags just improved readability for me, of the below line where we making the sampling decision : 
https://github.com/wavefrontHQ/java/pull/421/files#diff-6cf1decc90bf3c44cb59dcbb05ad9cb7R356

PR #433, opened 2019-08-16
Sanitize span name of Jaeger and Zipkin integration

haosong:


proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerThriftCollectorHandler.java
```diff
@@ -345,7 +346,7 @@ private void processSpan(io.jaegertracing.thriftjava.Span span,
     }
     Span wavefrontSpan = Span.newBuilder()
         .setCustomer("dummy")
-        .setName(span.getOperationName())
+        .setName(sanitizeValue(span.getOperationName()))
```
basilisk487: sanitizeValue returns a value wrapped in quotes so it's going to ruin the span name with adding extraneous quotes. We should fix sanitizeValue.

akodali18:
is this not needed anymore or we lost track ?

PR #434, opened 2019-08-19
Fix possibly incorrectly classified blocked points (rejected vs blocked)

basilisk487:


proxy/src/main/java/com/wavefront/agent/preprocessor/Preprocessor.java
```diff
@@ -56,6 +56,9 @@ public boolean filter(@Nonnull T item) {
    * @return true if all predicates returned "true"
    */
   public boolean filter(@Nonnull T item, @Nullable String[] messageHolder) {
+    if (messageHolder != null) {
```
sushantdewan123: add a comment?

PR #449, opened 2019-09-17
PUB-152: Support 'add' operation for sourceTags

basilisk487:


proxy/src/main/java/com/wavefront/agent/QueuedAgentService.java
```diff
@@ -881,10 +881,20 @@ public void execute(Object callback) {
       try {
         switch (messageType) {
           case tag:
-            if (actionType == ActionType.delete)
-              response = service.removeTag(id, token, tagValues[0]);
-            else
-              response = service.setTags(id, token, Arrays.asList(tagValues));
+            switch (actionType) {
+              case add:
+                response = service.appendTag(id, token, tagValues[0]);
```
davideauee: any reason why not to allow a loop through a set of new tags? i understand that the API does not support multiple additions at once but as i understand it, it is coming. In that context would it not be acceptable to support the feature here early. (once the API gets upgraded this code will have to be updated anyway due to the validation logic)

basilisk487: @davideauee Each `@SourceTag` operation is supposed to be atomic, and that's no longer the case if each tag is added using a separate API call. That makes error handling (like throttling due to rate limiting) and subsequent retries more challenging. We could split these multi-tag "add" (and "delete") requests into individual separate ones and handle them sequentially. In this case, it's theoretically possible that these tags would be processed out of order; and internal proxy metrics would show the actual number of source tag operations performed, not the number of `@SourceTag` lines sent by users. Would that be ok?

davideauee: I see your point about the atomicity issue.
your multi-add/delete suggestion sounds OK to me.

Ultimately the iteration will occur within the proxy or in our code. the difference being that if it is in your code then it can be optimized

PR #457, opened 2019-10-02
Add default preprocessor rule to Jaeger/Zipkin ports to sanitize spans

hanwavefront:


proxy/src/main/java/com/wavefront/agent/preprocessor/SpanSanitizeTransformer.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /**
+   * Sanitize a string so that every invalid character is replaced with a dash.
+   *
+   * @param s The string to sanitize.
+   * @return A {@link Pair} containing the resulting string and a boolean indicating whether any
+   * sanitization was applied.
+   */
+  private Pair<String, Boolean> sanitize(String s) {
```
sushantdewan123: if this is defined in tier 1 SDK, should we just use that instead of defining it again?

hanwavefront: the method in the SDK wraps the string in quotes, but i think doing that here would be incorrect.

proxy/src/main/java/com/wavefront/agent/preprocessor/SpanSanitizeTransformer.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+    StringBuilder sb = new StringBuilder();
+    boolean ruleApplied = false;
+    for (int i = 0; i < s.length(); i++) {
+      char cur = s.charAt(i);
+      boolean isLegal = true;
+      if (!(44 <= cur && cur <= 57) && !(65 <= cur && cur <= 90) && !(97 <= cur && cur <= 122) &&
+          cur != 95) {
+        if (!((i == 0 && cur == 0x2206) || (i == 0 && cur == 0x0394) || (i == 0 && cur == 126))) {
+          // first character can also be \u2206 (∆ - INCREMENT) or \u0394 (Δ - GREEK CAPITAL LETTER DELTA)
```
basilisk487: Spans can't start with a delta character

basilisk487: Also, we shouldn't replace all these characters in the span name as it makes a lot of people unhappy

hanwavefront: Removed the first-character conditions since those only apply to metric names.  This method only gets applied to source and tag keys, not span name.

proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -315,14 +317,32 @@ protected void startListeners() {
...
     portIterator(traceListenerPorts).forEachRemaining(strPort ->
         startTraceListener(strPort, handlerFactory, compositeSampler));
-    portIterator(traceJaegerListenerPorts).forEachRemaining(strPort ->
-        startTraceJaegerListener(strPort, handlerFactory,
-            new InternalProxyWavefrontClient(handlerFactory, strPort), compositeSampler));
+    portIterator(traceJaegerListenerPorts).forEachRemaining(strPort -> {
+      PreprocessorRuleMetrics ruleMetrics = new PreprocessorRuleMetrics(
+          Metrics.newCounter(new TaggedMetricName("spanSanitization", "count", "port", strPort)),
+          Metrics.newCounter(new TaggedMetricName("spanSanitization", "cpu_nanos", "port", strPort)),
+          Metrics.newCounter(new TaggedMetricName("spanSanitization", "checked-count", "port",
```
basilisk487: I think these two can (and should) be null. These metrics are intended to help users identify poorly performing preprocessor rules, and there's nothing user-configurable here, so it can only be more confusing. One of those cases where more metrics doesn't necessarily mean it's better:)

proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -315,14 +317,32 @@ protected void startListeners() {
 
     portIterator(traceListenerPorts).forEachRemaining(strPort ->
         startTraceListener(strPort, handlerFactory, compositeSampler));
-    portIterator(traceJaegerListenerPorts).forEachRemaining(strPort ->
-        startTraceJaegerListener(strPort, handlerFactory,
-            new InternalProxyWavefrontClient(handlerFactory, strPort), compositeSampler));
+    portIterator(traceJaegerListenerPorts).forEachRemaining(strPort -> {
+      PreprocessorRuleMetrics ruleMetrics = new PreprocessorRuleMetrics(
+          Metrics.newCounter(new TaggedMetricName("spanSanitization", "count", "port", strPort)),
```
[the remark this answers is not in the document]
hanwavefront: Updated

proxy/src/main/java/com/wavefront/agent/PushAgent.java
hanwavefront: Updated cpu nanos and checked count to be null.

proxy/src/main/java/com/wavefront/agent/preprocessor/SpanSanitizeTransformer.java
```diff
@@ -0,0 +1,73 @@
...
+    boolean ruleApplied = false;
+    for (int i = 0; i < s.length(); i++) {
+      char cur = s.charAt(i);
+      boolean isLegal = true;
+      if (!(44 <= cur && cur <= 57) && !(65 <= cur && cur <= 90) && !(97 <= cur && cur <= 122) &&
+          cur != 95) {
+        isLegal = false;
+        ruleApplied = true;
+      }
+      sb.append(isLegal ? cur : '-');
```
basilisk487: I think it would be more efficient to check if we need to do any replacements first rather than always rebuilding the string character by character, but it's not that big of a deal.

hanwavefront: Refactored a bit. Now it does a validation check before rebuilding the strings. And helper no longer has to return a Pair. We should probably consolidate the behavior between sdk, proxy, and javalib at some point.

PR #464, opened 2019-10-11
update open_source_licenses.txt for release 5.1

ajayj89:


ajayj89:
This is the case when license file was committed to release branch, but we need the same to be updated in master as well.
This process of committing the license file to both release branch and master will be automated.

PR #466, opened 2019-10-16
support mac proxy

djia-vm-wf:


vikramraman:
lgtm. Can merge once Vasily takes a look.

PR #468, opened 2019-10-22
support different blocked entity log

djia-vm-wf:
test locally on invalid point to 2878, invalid span to 30000, invalid histogram to 40000, non-delta counter to 50000

- different loggers in config.ini, all loggers enabled in log4j2 -> different log files
- no logger in config.ini, all logger enabled in log4j2 -> all in blockedPoints
- span/histogram logger in config,  all loggers enable in log4j2 -> span/histogram log file separate, others go to blockedPoints
- span/histogram logger in config,  corresponding enable in log4j2 -> span/histogram log file only
- No logger in config, no log4j2 -> no log file, console is fine

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
+++ b/proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
@@ -200,18 +200,17 @@ public abstract class AbstractAgent {
       " to 5.")
   protected Integer pushBlockedSamples = 5;
 
-  @Parameter(names = {"--blockedPointsLoggerName"}, description = "Logger Name for blocked points")
+  @Parameter(names = {"--blockedPointsLoggerName"}, description = "Logger Name for blocked " +
+      "points. " + "Default: RawBlockedPoints")
   protected String blockedPointsLoggerName = "RawBlockedPoints";
-  protected Logger blockedPointsLogger;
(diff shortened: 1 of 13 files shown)


proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
```diff
@@ -200,6 +200,19 @@
       " to 5.")
   protected Integer pushBlockedSamples = 5;
 
+  @Parameter(names = {"--blockedPointsLoggerName"}, description = "Logger Name for blocked points")
+  protected String blockedPointsLoggerName = "RawBlockedPoints";
+  protected Logger blockedPointsLogger;
```
basilisk487: Let's not mix settings and properties and move loggers to PushAgent

proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
```diff
@@ -200,6 +200,19 @@
       " to 5.")
   protected Integer pushBlockedSamples = 5;
 
+  @Parameter(names = {"--blockedPointsLoggerName"}, description = "Logger Name for blocked points")
```
basilisk487: Specify default value in description

PR #474, opened 2019-11-16
Add option to receive Jaeger spans via HTTP instead of TChannel

hanwavefront:


proxy/src/test/java/com/wavefront/agent/listeners/tracing/JaegerPortUnificationHandlerTest.java
```diff
@@ -0,0 +1,195 @@
...
+import wavefront.report.SpanLogs;
+
+import static org.easymock.EasyMock.createNiceMock;
+import static org.easymock.EasyMock.expect;
+import static org.easymock.EasyMock.expectLastCall;
+import static org.easymock.EasyMock.replay;
+import static org.easymock.EasyMock.reset;
+import static org.easymock.EasyMock.verify;
+
+/**
```
srujann: add a brief description of what these tests are doing?

proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
```diff
@@ -503,8 +503,12 @@
       "data. Defaults to none.")
   protected String traceListenerPorts;
 
+  @Parameter(names = {"--traceJaegerListenerProtocol"}, description = "Protocol used to listen " +
```
akodali18: nit: IMO we can name this something shorter like "traceJaegerProtocol"

proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -325,16 +326,34 @@ protected void startListeners() {
...
+            Metrics.newCounter(new TaggedMetricName("point.spanSanitize", "count", "port", strPort)),
+            null, null
+        );
+        preprocessors.getSystemPreprocessor(strPort).forSpan().addTransformer(
+            new SpanSanitizeTransformer(ruleMetrics));
+        startTraceJaegerHttpListener(strPort, handlerFactory,
+            new InternalProxyWavefrontClient(handlerFactory, strPort), compositeSampler);
+      });
+    } else {
+      logger.severe("Unable to start Jaeger listener due to invalid listener protocol: " +
```
akodali18: should we just log a warn message and default to use jaeger with tchannel listener instead in this case ?

proxy/src/test/java/com/wavefront/agent/listeners/tracing/JaegerPortUnificationHandlerTest.java
```diff
@@ -0,0 +1,195 @@
...
+  private final static String DEFAULT_SOURCE = "jaeger";
+  private ReportableEntityHandler<Span> mockTraceHandler =
+      MockReportableEntityHandlerFactory.getMockTraceHandler();
+  private ReportableEntityHandler<SpanLogs> mockTraceSpanLogsHandler =
+      MockReportableEntityHandlerFactory.getMockTraceSpanLogsHandler();
+  private ChannelHandlerContext mockCtx = createNiceMock(ChannelHandlerContext.class);
+
+  private long startTime = System.currentTimeMillis();
+
+  @Test
```
akodali18: nit: We need to reformat the unit tests at https://github.com/wavefrontHQ/wavefront-proxy/blob/master/proxy/src/test/java/com/wavefront/agent/listeners/tracing/JaegerThriftCollectorHandlerTest.java iI guess since we moved some methods to a util function. (something for future, for now its good).

proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerThriftCollectorHandler.java
```diff
@@ -185,7 +147,10 @@ public JaegerThriftCollectorHandler(String handle,
       ThriftRequest<Collector.submitBatches_args> request) {
     for (Batch batch : request.getBody(Collector.submitBatches_args.class).getBatches()) {
       try {
-        processBatch(batch);
+        processBatch(batch, null, DEFAULT_SOURCE, proxyLevelApplicationName, spanHandler,
```
akodali18: Post this change, I think this class name is deceiving i.e. "JaegerThriftCollectorHandler" since we handle thrift formatted data over tchannel and http both. Should we rename it to "JaegerTChannelCollectorHandler"  ?

proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
basilisk487: agreed, let's split into `traceJaegerListenerPorts` (tchannel) and `traceJaegerHttpListenerPorts`

proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerPortUnificationHandler.java
```diff
@@ -0,0 +1,218 @@
...
+                               ReportableEntityHandler<SpanLogs> spanLogsHandler,
+                               @Nullable WavefrontSender wfSender,
+                               Supplier<Boolean> traceDisabled,
+                               Supplier<Boolean> spanLogsDisabled,
+                               @Nullable Supplier<ReportableEntityPreprocessor> preprocessor,
+                               Sampler sampler,
+                               boolean alwaysSampleErrors,
+                               @Nullable String traceJaegerApplicationName,
+                               Set<String> traceDerivedCustomTagKeys) {
+    super(TokenAuthenticatorBuilder.create().build(), healthCheckManager, handle);
```
basilisk487: Since this is http, it's really easy to add true auth support to this port as well, just pass the TokenAuthenticator as an argument from PushAgent.

proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
hanwavefront: Do we want to support a different application via HTTP too? In which case we would add traceJaegerHttpApplicationName as well.  Or even multiple applications, in which case we could add a property like `traceJaegerConfig=14267:tchannel:App1,14268:http:App2`

akodali18: If i understood your question correctly, Jaeger spans sent via any method(http/tchannel) should have "Application" as one of the span tags to distinguish between applications. 
So It is just two ports traceJaegerListenerPorts (tchannel) and traceJaegerHttpListenerPorts (http) as vasily mentioned.

hanwavefront: yes, and I believe that tag value is pulled from proxy-level Jaeger application name in the config (`traceJaegerApplicationName`), so I'm wondering if we need to expand the config to allow for multiple application names (per protocol or per port)?

akodali18: If customer has multiple applications best would be to ask them to use "Application" tag as a Jaeger process level tag rather than depending on the Proxy level tags.
Don't think we need to handle this case at the proxy level.

hanwavefront: ah i see. thanks for clarifying

proxy/src/test/java/com/wavefront/agent/listeners/tracing/JaegerPortUnificationHandlerTest.java
hanwavefront: added a generic description

proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
hanwavefront: added `traceJaegerHttpListenerPorts` and removed `traceJaegerProtocol`

proxy/src/main/java/com/wavefront/agent/PushAgent.java
hanwavefront: no longer applicable

proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerThriftCollectorHandler.java
hanwavefront: renamed it to `JaegerTChannelCollectorHandler`

proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerPortUnificationHandler.java
hanwavefront: updated to pass in the TokenAuthenticator

proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -535,7 +527,7 @@ protected void startTraceJaegerListener(String strPort,
             build();
         server.
             makeSubChannel("jaeger-collector", Connection.Direction.IN).
-            register("Collector::submitBatches", new JaegerThriftCollectorHandler(strPort,
+            register("Collector::submitBatches", new JaegerTChannelCollectorHandler(strPort,
```
akodali18: nit: modify  "listener-jaeger-thrift-" -> "listener-jaeger-tchannel-"

hanwavefront: updated

PR #476, opened 2019-11-25
Sanitize span name and tag value for Jaeger/Zipkin; use internal repo…

hanwavefront:
…rter v1.1

Continuation of https://github.com/wavefrontHQ/wavefront-proxy/pull/433/ to make Jaeger/Zipkin span sanitization consistent with SDKs.

- updated span preprocessor rule to sanitize span name and tag values
- replace '*' with '-' in span name
- i don't think quotes need to be escaped in the preprocessor because SpanSerializer will escape them
- derived metrics are sanitized by the internal reporter's sender, so updated to use latest version of the internal reporter

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/preprocessor/SpanSanitizeTransformer.java
+++ b/proxy/src/main/java/com/wavefront/agent/preprocessor/SpanSanitizeTransformer.java
@@ -1,9 +1,10 @@
 package com.wavefront.agent.preprocessor;
 
 import com.google.common.base.Function;
-import org.checkerframework.checker.nullness.qual.Nullable;
 import wavefront.report.Annotation;
 import wavefront.report.Span;
+
+import javax.annotation.Nonnull;
 
 /**
  * Sanitize spans (e.g., span source and tag keys) according to the same rules that are applied at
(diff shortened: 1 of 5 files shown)


proxy/src/main/java/com/wavefront/agent/preprocessor/SpanSanitizeTransformer.java
```diff
@@ -18,59 +19,79 @@ public SpanSanitizeTransformer(final PreprocessorRuleMetrics ruleMetrics) {
...
-
-  private boolean characterIsValid(char c) {
-    // Legal characters are 44-57 (,-./ and numbers), 65-90 (upper), 97-122 (lower), 95 (_)
-    return (44 <= c && c <= 57) || (65 <= c && c <= 90) || (97 <= c && c <= 122) || c == 95;
-  }
-
   /**
    * Sanitize a string so that every invalid character is replaced with a dash.
    */
   private String sanitize(String s) {
```
sushantdewan123: Do you want to make Java SDK as source of truth for sanitize() and sanitizeValue() methods?

hanwavefront:
> You should also sanitize() using tier 1 SDK
> https://github.com/wavefrontHQ/wavefront-proxy/blob/master/proxy/src/main/java/com/wavefront/agent/listeners/tracing/SpanDerivedMetricsUtils.java#L129

I think this should be fine because it'll get sanitized by the tier 1 SDK when the internal reporter sends it.  The current sanitization is consistent with the derived metric sanitization in the OpenTracing SDK: https://github.com/wavefrontHQ/wavefront-opentracing-sdk-java/blob/master/src/main/java/com/wavefront/opentracing/WavefrontTracer.java#L326

proxy/src/main/java/com/wavefront/agent/preprocessor/SpanSanitizeTransformer.java
hanwavefront: Ideally yes, will need to open a PR on the SDK first because sanitize() currently has behavior that pertains only to deltas while sanitizeValue() wraps the string in quotes.

hanwavefront:
> > You should also sanitize() using tier 1 SDK
> > https://github.com/wavefrontHQ/wavefront-proxy/blob/master/proxy/src/main/java/com/wavefront/agent/listeners/tracing/SpanDerivedMetricsUtils.java#L129
> 
> I think this should be fine because it'll get sanitized by the tier 1 SDK when the internal reporter sends it. The current sanitization is consistent with the derived metric sanitization in the OpenTracing SDK: https://github.com/wavefrontHQ/wavefront-opentracing-sdk-java/blob/master/src/main/java/com/wavefront/opentracing/WavefrontTracer.java#L326

@sushantdewan123 Updated sanitize() to mirror the SDK, minus wrapping in quotes.

proxy/src/main/java/com/wavefront/agent/preprocessor/SpanSanitizeTransformer.java
hanwavefront: Updated to use `sanitizeWithoutQuotes` from the latest version of the SDK

PR #478, opened 2019-12-12
Proxy refactoring 

basilisk487:
- Complete queue pipeline refactoring: instead of a single queue for the entire proxy,  each port + data type combination has its own queue
- Update tape library to 2.0 that no longer has the 1GB per file limitation
- To reduce contention, retryThreads is no longer configurable separately, there is 1 to 1 affinity between senders and queues; and timing priority is controlled by a master thread that gives queues that have the “oldest” data a boost. 
- Separate configurable rate limits for points/histograms/sourceTags/events/spans/spanLogs
- Config properties refactored out of AbstractAgent into a separate class
- Code de-duping and cruft cleanup 
- Unit test coverage improved


proxy/src/main/java/com/wavefront/agent/ProxyCheckinScheduler.java
```diff
@@ -0,0 +1,264 @@
...
+          if (succesfulCheckins.get() == 0) {
+            logger.warning("Aborting start-up");
+            System.exit(-5);
+          }
+          break;
+        case 407:
+          checkinError("HTTP 407 Proxy Authentication Required: Please verify that " +
+              "proxyUser and proxyPassword settings are correct and make sure your HTTP proxy" +
+              " is not rate limiting!");
+          break;
```
MikeMcMahon: Candidate for 407 since changing proxyUser and proxyPassword requires restart anyhow (like 405 which is likely a misconfiguration)

proxy/src/main/java/com/wavefront/agent/ProxyUtil.java
```diff
@@ -0,0 +1,189 @@
...
+        } catch (IOException e) {
+          logger.log(Level.SEVERE, "Cannot read from " + proxyIdFile, e);
+          System.exit(1);
+        }
+      } else {
+        logger.severe(proxyIdFile + " is not a file!");
+        System.exit(1);
+      }
+    } else {
+      logger.info("Proxy Id created: " + proxyId);
```
MikeMcMahon: would this make more sense after the try/catch since each startup would potentially then create a new proxyId (unless the file was manually created, and this is intended as a means to give the user a way to bypass failing permissions??)

proxy/src/main/java/com/wavefront/agent/ProxyUtil.java
```diff
@@ -0,0 +1,189 @@
...
+   * are much higher, so we use a random 32bit hex string instead.
+   *
+   * @return unique process identifier string
+   */
+  static String getProcessId() {
+    try {
+      final String runtime = ManagementFactory.getRuntimeMXBean().getName();
+      if (runtime.indexOf("@") >= 1) {
+        long id = Long.parseLong(runtime.substring(0, runtime.indexOf("@")));
+        if (id > 1000) {
```
MikeMcMahon: 1000 is arbitrary, since pids are assigned based on request, ofc in a containerized environment the pid will almost always be 1 as the proxy will be the first and basically only thing running on the container (or should be).  Other environments it's a mixed bag based on the OS and what is running.  If collisions are a problem maybe the pid isn't best choice and best choice is just the random hex below.  1 and 0 are reserved (most of the time) and anything after that is fair game, though most of the low numbers will be consumed by low level kernel processes and other system services that start well before other services.

proxy/src/main/java/com/wavefront/agent/channel/ChannelUtils.java
```diff
@@ -97,6 +82,53 @@ public static void writeHttpResponse(final ChannelHandlerContext ctx,
...
+                                       boolean keepAlive) {
+    // Decide whether to close the connection or not.
+    if (keepAlive) {
+      // Add keep alive header as per:
+      // - http://www.w3.org/Protocols/HTTP/1.1/draft-ietf-http-v11-spec-01.html#Connection
+      response.headers().set(HttpHeaderNames.CONNECTION, HttpHeaderValues.KEEP_ALIVE);
+      ctx.write(response);
+    } else {
+      ctx.write(response);
+      ctx.writeAndFlush(Unpooled.EMPTY_BUFFER).addListener(ChannelFutureListener.CLOSE);
```
MikeMcMahon: can this just be collapsed into a single writeAndFlush() call?

proxy/src/main/java/com/wavefront/agent/queueing/RetryTaskConverter.java
```diff
@@ -0,0 +1,123 @@
...
+            }
+            return (T) objectMapper.readValue(stream, DataSubmissionTask.class);
+          } catch (Throwable t) {
+            logger.warning("Unable to restore persisted task: " + t);
+          }
+        } else {
+          logger.warning("Unable to restore persisted task - corrupted header, ignoring");
+        }
+      } else {
+        logger.warning("Unable to restore persisted task - corrupted header, ignoring");
```
MikeMcMahon: wouldn't this be more specific to say missing header as the bytes to read would be 0?

basilisk487: Agreed

proxy/src/main/java/com/wavefront/agent/ProxyUtil.java
basilisk487: useless trivia fact: in container envs with our proxy image the pid is almost always 15 (occasionally it's off by 1).
but in general i agree, let's stick to random hex.

proxy/src/main/java/com/wavefront/agent/ProxyUtil.java
basilisk487: By default we use ephemeral mode which generates new proxy id at each startup, it just doesn't persist it in a file. A new proxy ID is only valid if it's been associated with a valid token with proper permissions.

proxy/src/main/java/com/wavefront/agent/channel/ChannelUtils.java
basilisk487: yup.

proxy/src/main/java/com/wavefront/agent/ProxyUtil.java
```diff
@@ -0,0 +1,165 @@
...
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelInitializer;
+import io.netty.channel.ChannelPipeline;
+import io.netty.channel.socket.SocketChannel;
+import io.netty.handler.timeout.IdleStateHandler;
+
+import javax.annotation.Nullable;
+import java.io.File;
+import java.io.IOException;
+import java.lang.management.ManagementFactory;
```
kentwang929: unused import

basilisk487: fixed

proxy/src/main/java/com/wavefront/agent/config/Configuration.java
```diff
@@ -15,14 +15,15 @@ protected void ensure(boolean b, String message) throws ConfigurationException {
 
   public abstract void verifyAndInit() throws ConfigurationException;
 
-  private static ObjectMapper objectMapper = new ObjectMapper();
+  private static final ObjectMapper objectMapper = new ObjectMapper();
```
panghy: OBJECT_MAPPER

proxy/src/main/java/com/wavefront/agent/config/Configuration.java
```diff
@@ -15,14 +15,15 @@ protected void ensure(boolean b, String message) throws ConfigurationException {
...
 
   @Override
   public String toString() {
     try {
       return objectMapper.writeValueAsString(this);
     } catch (JsonProcessingException e) {
-      return super.toString();
+      throw new RuntimeException(e);
+      //return super.toString();
```
panghy: remove

proxy/src/main/java/com/wavefront/agent/handlers/SourceTagSenderTask.java
```diff
@@ -0,0 +1,118 @@
...
+            default:
+          }
+        } else {
+          final List<SourceTag> remainingItems = new ArrayList<>();
+          iterator.forEachRemaining(remainingItems::add);
+          undoBatch(remainingItems);
+          // if proxy rate limit exceeded, try again in 1/4..1/2 of flush interval
+          // to introduce some degree of fairness.
+          nextRunMillis = (int) (1 + Math.random()) * nextRunMillis / 4;
+          throttledLogger.info("[" + handlerKey.getHandle() + " thread " + threadId +
```
panghy: in cases like this, can we use a function instead? maybe not for this CR but constructing these strings and evaluating Math.random() for something that could potentially be dropped due to throttling is just plain inefficient

proxy/src/main/java/com/wavefront/agent/listeners/WriteHttpJsonPortUnificationHandler.java
```diff
@@ -161,7 +153,7 @@ private void reportMetrics(JsonNode metrics) {
         } else {
           builder.setValue(value.asLong());
         }
-        List<ReportPoint> parsedPoints = Lists.newArrayListWithExpectedSize(1);
+        List<ReportPoint> parsedPoints = Lists.newArrayListWithCapacity(1);
```
panghy: can just do new ArrayList<>(1) these days

panghy: or just use List.of()

proxy/src/main/java/com/wavefront/agent/preprocessor/PreprocessorUtil.java
```diff
@@ -83,19 +82,36 @@ public static String expandPlaceholders(String input, @Nonnull Span span) {
...
+    switch (actionSubtype) {
+      case TRUNCATE:
+        return input.substring(0, maxLength);
+      case TRUNCATE_WITH_ELLIPSIS:
+        return input.substring(0, maxLength - 3) + "...";
+      default:
+        return input;
+    }
+  }
+
```
panghy: blank line

proxy/src/main/java/com/wavefront/agent/preprocessor/PreprocessorUtil.java
```diff
@@ -83,19 +82,36 @@ public static String expandPlaceholders(String input, @Nonnull Span span) {
...
+   * @return truncated string
+   */
+  public static String truncate(String input, int maxLength, LengthLimitActionType actionSubtype) {
+    switch (actionSubtype) {
+      case TRUNCATE:
+        return input.substring(0, maxLength);
+      case TRUNCATE_WITH_ELLIPSIS:
+        return input.substring(0, maxLength - 3) + "...";
+      default:
+        return input;
```
panghy: throw

proxy/src/main/java/com/wavefront/agent/data/EntityProperties.java
```diff
@@ -0,0 +1,147 @@
...
+   * @return true if we should split on pushback
+   */
+  boolean isSplitPushWhenRateLimited();
+
+  /**
+   * Get base in seconds for retry thread exponential backoff.
+   *
+   * @return exponential backoff base value
+   */
+  double getRetryBackoffBaseSeconds();
```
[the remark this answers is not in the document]
basilisk487: The idea was to support overrides to initially configured values when these parameters are controlled by the backend; and setting it to null would make it revert to the original value:
```
    @Override
    public int getItemsPerBatch() {
      return ObjectUtils.firstNonNull(itemsPerBatch, getItemsPerBatchOriginal());
    }

    @Override
    public void setItemsPerBatch(@Nullable Integer itemsPerBatch) {
      this.itemsPerBatch = itemsPerBatch;
    }
```

 It felt that using `null` to revert to the original value was a more elegant approach than creating separate "revert" methods. Also, this approach is in line with AgentConfiguration, where properties become null when these parameters are no longer controlled by the backend.

proxy/src/main/java/com/wavefront/agent/handlers/SourceTagSenderTask.java
basilisk487: Fixed by switching to Supplier<String> instead.

PR #479, opened 2019-12-30
custom tracing port

djia-vm-wf:


proxy/src/main/java/com/wavefront/agent/listeners/tracing/CustomTracingPortUnificationHandler.java
```diff
@@ -0,0 +1,239 @@
...
+        preprocessor, handlerFactory.getHandler(HandlerKey.of(ReportableEntityType.TRACE, handle)),
+        sampler, alwaysSampleErrors, traceDisabled);
+  }
+
+  @VisibleForTesting
+  public CustomTracingPortUnificationHandler(
+      final String handle, final TokenAuthenticator tokenAuthenticator,
+      final HealthCheckManager healthCheckManager,
+      @Nullable WavefrontSender wfSender,
+      final ReportableEntityDecoder<String, Span> traceDecoder,
```
basilisk487: Need to bring back support for SpanLogs that was taken out

proxy/src/main/java/com/wavefront/agent/listeners/tracing/CustomTracingPortUnificationHandler.java
```diff
@@ -0,0 +1,239 @@
...
+import static com.wavefront.sdk.common.Constants.SERVICE_TAG_KEY;
+import static com.wavefront.sdk.common.Constants.SHARD_TAG_KEY;
+
+/**
+ * Handler that process trace data sent from tier 1 SDK.
+ *
+ * @author djia@vmware.com
+ */
+@ChannelHandler.Sharable
+public class CustomTracingPortUnificationHandler extends AbstractLineDelimitedHandler {
```
basilisk487: It would be a lot more efficient to extend TracePortUnificationHandler since they share a lot of common code. Extract parts that are different into a separate method and override just that.

proxy/src/main/java/com/wavefront/agent/listeners/tracing/CustomTracingPortUnificationHandler.java
```diff
@@ -0,0 +1,239 @@
...
+              case ERROR_SPAN_TAG_KEY:
+                isError = annotation.getValue();
+                break;
+            }
+          }
+
+          if (applicationName.equals(NULL_TAG_VAL) || serviceName.equals(NULL_TAG_VAL)) {
+            logger.warning("Ingested spans discarded because span application/service name is " +
+                "missing.");
+            discardedSpans.inc();
```
srujann: seems like the span is already reported above, so this code is not doing what you intended.

basilisk487:
Also, needs unit tests (now every pipeline has its own unit tests). It's possible to do that without any thread sleeps - take a look at how delta counter aggregator is tested now.

proxy/src/main/java/com/wavefront/agent/listeners/tracing/CustomTracingPortUnificationHandler.java
```diff
@@ -0,0 +1,239 @@
...
+            return;
+          }
+
+          discoveredHeartbeatMetrics.putIfAbsent(reportWavefrontGeneratedData(wfInternalReporter,
+              object.getName(), applicationName, serviceName, cluster, shard, object.getSource(),
+              componentTagValue, Boolean.parseBoolean(isError), object.getDuration(),  new HashSet<>(),
```
basilisk487: do we need to create new HashSet<>() for every span?

srujann: Looking at Jaeger and Zipkin handler, seems like this should be passed in through a config instead of empty set.

proxy/src/main/java/com/wavefront/agent/listeners/tracing/TracePortUnificationHandler.java
```diff
@@ -174,13 +174,17 @@ protected void processLine(final ChannelHandlerContext ctx, @Nonnull String mess
...
           t -> t.getKey().equals(ERROR_SPAN_TAG_KEY) && t.getValue().equals(ERROR_SPAN_TAG_VAL));
-      if (sampleError || sample(object)) {
-        handler.report(object);
-      }
+      report(object, sampleError);
+    }
+  }
+
+  protected void report(Span object, boolean sampleError) {
+    if (sampleError || sample(object)) {
```
srujann: you can have this outside of the report method, report method do the work of only reporting. Also add a java doc for this method since its exposed to extending members.

proxy/src/main/java/com/wavefront/agent/listeners/tracing/CustomTracingPortUnificationHandler.java
```diff
@@ -0,0 +1,173 @@
...
+
+      if (applicationName.equals(NULL_TAG_VAL) || serviceName.equals(NULL_TAG_VAL)) {
+        logger.warning("Ingested spans discarded because span application/service name is " +
+            "missing.");
+        discardedSpans.inc();
+        return;
+      }
+
+      if (sampleError || sample(object)) {
+        handler.report(object);
```
basilisk487: We'll never get here if wfInternalReporter == null

basilisk487: I believe the right approach would be to only disable internal metrics if wfInternalReporter is null, but we still want to validate spans (check for presence of application/service tags) and report them.

proxy/src/main/java/com/wavefront/agent/listeners/tracing/CustomTracingPortUnificationHandler.java
```diff
@@ -0,0 +1,173 @@
...
+      Supplier<Boolean> spanLogsDisabled, @Nullable WavefrontSender wfSender,
+      Set<String> traceDerivedCustomTagKeys) {
+    super(handle, tokenAuthenticator, healthCheckManager, traceDecoder, spanLogsDecoder,
+        preprocessor, handler, spanLogsHandler, sampler, alwaysSampleErrors, traceDisabled, spanLogsDisabled);
+    this.wfSender = wfSender;
+    this.discoveredHeartbeatMetrics = new ConcurrentHashMap<>();
+    this.traceDerivedCustomTagKeys = traceDerivedCustomTagKeys;
+
+    if (wfSender != null) {
+      wfInternalReporter = new WavefrontInternalReporter.Builder().
```
basilisk487: Might be a good idea to actually pass a WavefrontInternalReporter as a constructor argument (as well as starting it outside of the constructor). The way it's implemented right now it's quite difficult to create a mock for the reporter for testing purposes.

Issue #483, opened 2020-01-14
Support white-list restriction on span tag keys


panghy:
As titled, only a set of span tag keys would be allowed (some fields should always be allowed such as service, application, host, cluster and shard).

PR #485, opened 2020-01-16
#483 Support whitelist restriction on span tag keys

basilisk487:


proxy/src/main/java/com/wavefront/agent/preprocessor/SpanWhitelistAnnotationTransformer.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  @Nullable
+  @Override
+  public Span apply(@Nullable Span span) {
+    if (span == null) return null;
+    long startNanos = ruleMetrics.ruleStart();
+    List<Annotation> annotations = span.getAnnotations().stream().
+        filter(x -> {
+          if (!whitelistedKeys.containsKey(x.getKey())) return false;
```
panghy: technically you can express this as two filters for slightly easier readability

basilisk487: @panghy fixed, this should be more readable

PR #492, opened 2020-01-30
Fix duplicate Authorization headers

basilisk487:


proxy/src/main/java/com/wavefront/agent/api/APIContainer.java
```diff
@@ -158,9 +158,10 @@ private ResteasyProviderFactory createProviderFactory() {
     }
     factory.register(AcceptEncodingGZIPFilter.class);
     factory.register((ClientRequestFilter) context -> {
-          if (context.getUri().getPath().contains("/v2/wfproxy") ||
+          if ((context.getUri().getPath().contains("/v2/wfproxy") ||
```
sushantdewan123: can you add a quick comment before the if statement?

PR #498, opened 2020-02-18
Handle data from DataDog of type:rate

conorbev:
When DataDog sends statsd counter data the JSON contains a value that when displayed by DataDog is multiplied by the "interval" property supplied in the JSON.

This change allows Wavefront to store and render that same value. The modified value also matches what the telegraf statsd input plugin sends for the same DogStatsD data so this seems to be a quirk of how DataDog sends and parses the data.

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/listeners/DataDogPortUnificationHandler.java
+++ b/proxy/src/main/java/com/wavefront/agent/listeners/DataDogPortUnificationHandler.java
@@ -302,6 +302,10 @@ public class DataDogPortUnificationHandler extends AbstractHttpOnlyHandler {
         tags.putAll(systemTags);
       }
       extractTags(tagsNode, tags); // tags sent with the data override system host-level tags
+      JsonNode deviceNode = metric.get("device"); // Include a device= tag on the data if that property exists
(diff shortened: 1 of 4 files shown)


proxy/src/main/java/com/wavefront/agent/listeners/DataDogPortUnificationHandler.java
```diff
@@ -306,6 +306,13 @@ private boolean reportMetric(final JsonNode metric, @Nullable final AtomicIntege
       if (deviceNode != null) {
         tags.put("device", deviceNode.textValue());
       }
+      int interval = 1; // If the metric is of type rate its value needs to be multiplied by the specified interval
+      JsonNode type = metric.get("type");
+      if (type != null) {
+        if (type.textValue().equals("rate")) {
+          interval = metric.get("interval") == null ? 1 : metric.get("interval").intValue();
```
basilisk487: We probably want to handle bad data scenarios where "interval" field is not numeric

conorbev: ok, done

PR #500, opened 2020-02-20
Promote span.kind as RED metrics

yzheqing:
1. This PR allows the custom tag filter `span.kind`, to appear on P0/P1 application status page. The dropdown menu under `span.kind` contains the value of `client`, `server` or `unknown` (by default).
2. Tested with Jaeger/Zipkin integration: span data being sent through wavefront proxy can be filtered on P0/P1 through `span.kind` dropdown menu.


proxy/config.ini
```diff
@@ -12,3 +12,5 @@ pushLogLevel=SUMMARY
 pushValidationLevel=NUMERIC_ONLY
 
 idFile=wavefront_test_id
+
+traceDerivedCustomTagKeys=span.kind
```
hanwavefront: Users will have different config files, so instead of adding it here, let's programmatically include `span.kind` when handling `traceDerivedCustomTagKeys` in the code.

PR #501, opened 2020-02-21
PUB-211: Fix histogram.points.received metric not reporting (6.0 fix)

basilisk487:


proxy/src/main/java/com/wavefront/agent/handlers/AbstractReportableEntityHandler.java
```diff
@@ -63,57 +64,54 @@
...
     this.blockedItemsLogger = blockedItemsLogger;
 
-    MetricsRegistry registry = setupMetrics ? Metrics.defaultRegistry() : LOCAL_REGISTRY;
+    MetricsRegistry registry = reportReceivedStats ? Metrics.defaultRegistry() : LOCAL_REGISTRY;
     String metricPrefix = handlerKey.toString();
     MetricName receivedMetricName = new MetricName(metricPrefix, "", "received");
     MetricName deliveredMetricName = new MetricName(metricPrefix, "", "delivered");
     this.receivedCounter = registry.newCounter(receivedMetricName);
-    this.attemptedCounter = registry.newCounter(new MetricName(metricPrefix, "", "sent"));
+    this.attemptedCounter = Metrics.newCounter(new MetricName(metricPrefix, "", "sent"));
```
sushantdewan123: should this be registry.newCounter() instead of Metrics.newCounter() ?

basilisk487: @sushantdewan123 We want to always report .sent metric, but only report .received when reportReceivedCounter = true

PR #503, opened 2020-02-24
PUB-214 Support multi port or global(applies to all ports) preprocessor rules

akodali18:
Addresses https://github.com/wavefrontHQ/wavefront-proxy/issues/245


proxy/src/main/java/com/wavefront/agent/preprocessor/PreprocessorConfigManager.java
```diff
@@ -435,4 +453,8 @@ void loadFromStream(InputStream stream) {
       this.userPreprocessors = portMap;
     }
   }
+
+  private void validateAndAddTransformers(String strPort) {
+
```
basilisk487: empty private method? :)

akodali18: oops sorry forgot to remove.

PR #509, opened 2020-03-14
PUB-218 enhanced preprocessor rules.

akodali18:
Gist:
We introduced a new argument denoted in yaml with the key "if", whose value denotes a predicate function. The preprocessor rule will apply to a given entity if this predicate function evaluates to true. We will be calling this as the "v2PredicateKey" and its corresponding predicate function as "v2Predicate".

Semantics and behavior changes:
1. v2Predicate applies to all scopes except for "pointline"
2. For "Filtering" preprocessor rules, Previously v1Predicates(i.e. [match], [scope] arguments) were considered mandatory. With this change, v1Predicates are considered optional, if v2Predicate present. However, note that v1Predicates cannot be specified and left blank(this is still considered invalid). If both v1Predicates and v2Predicate is specified, then both need to be satisfied to apply the rule.


proxy/src/main/java/com/wavefront/agent/preprocessor/ReportPointAddTagIfNotExistsTransformer.java
```diff
@@ -11,17 +13,25 @@
  */
 public class ReportPointAddTagIfNotExistsTransformer extends ReportPointAddTagTransformer {
 
+  @Nullable
+  private final Map<String, Object> v2Predicate;
+
   public ReportPointAddTagIfNotExistsTransformer(final String tag,
                                                  final String value,
+                                                 @Nullable final Map<String, Object> v2Predicate,
```
basilisk487: I think this could be another performance issue, as you are basically parsing your predicate for every point. A better approach would be to parse the entire predicate tree from your Map<String, Object> into Predicate<T> and pass that as an argument. That would also allow you to cache compiled regex patterns, etc.

proxy/src/main/java/com/wavefront/agent/preprocessor/predicate/ContainsPredicate.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class ContainsPredicate<T> extends ComparisonPredicate<T>{
+
+  public ContainsPredicate(String scope, String value) {
+    super(scope, value);
+  }
+
+  @Override
+  public boolean test(T t) {
+    if (t instanceof ReportPoint) {
```
basilisk487: While this approach would work, using raw types and checking `instanceof` in a critical path (especially considering that the number of times test(T) is called is going to be many times more than the incoming data rate) is less than ideal in our case. Sometimes it's difficult to work around it and you are stuck with dealing with generics wildcards, but in this case it should be fairly easy - just implement `ReportPointContainsPredicate extends ComparisonPredicate<ReportPoint>` and `SpanContainsPredicate extends ComparisonPredicate<Span>` separately. As you are building your predicate tree you are already aware of the rule context, whether it's ReportPoint or Span, so that should be easier. While it does look like more code, it will be a more performant and scalable solution, and it also gives you flexibility to implement things like allContains/allStartWith/etc that would be applicable only to spans (or any possible future entity where annotations are a multimap).

proxy/src/main/java/com/wavefront/agent/preprocessor/PreprocessorUtil.java
```diff
@@ -140,4 +157,135 @@ public static int getInteger(Map<String, Object> ruleMap, String key, int defaul
...
+      case "metricName": return point.getMetric();
+      case "sourceName": return point.getHost();
+      default: return point.getAnnotations().get(scope);
+    }
+  }
+
+  public static List<String> getReportableEntityComparableValue(Object scope, Span span) {
+    switch ((String) scope) {
+      case "spanName": return Arrays.asList(span.getName());
+      case "sourceName": return Arrays.asList(span.getSource());
```
basilisk487: nit: use ImmutableList.of()

proxy/src/main/java/com/wavefront/agent/preprocessor/ReportPointAddTagTransformer.java
```diff
@@ -17,23 +20,29 @@
...
                                       final String value,
+                                      @Nullable final Map<String, Object> v2Predicate,
                                       final PreprocessorRuleMetrics ruleMetrics) {
     this.tag = Preconditions.checkNotNull(tag, "[tag] can't be null");
     this.value = Preconditions.checkNotNull(value, "[value] can't be null");
     Preconditions.checkArgument(!tag.isEmpty(), "[tag] can't be blank");
     Preconditions.checkArgument(!value.isEmpty(), "[value] can't be blank");
     Preconditions.checkNotNull(ruleMetrics, "PreprocessorRuleMetrics can't be null");
     this.ruleMetrics = ruleMetrics;
+    this.v2Predicate = PreprocessorUtil.parsePredicate(v2Predicate);
```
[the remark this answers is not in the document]
akodali18: thats a good idea, will do.

proxy/src/main/java/com/wavefront/agent/preprocessor/ReportPointAddTagTransformer.java
```diff
@@ -17,23 +20,29 @@
...
   }
 
   @Nullable
   @Override
   public ReportPoint apply(@Nullable ReportPoint reportPoint) {
     if (reportPoint == null) return null;
     long startNanos = ruleMetrics.ruleStart();
+
+    if (!v2Predicate.test(reportPoint)) return reportPoint;
+
```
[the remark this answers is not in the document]
akodali18: yes, i realized that and was thinking of adding them as a separate follow up change to not cloud this review. Added try/finally to all rules anyways.

akodali18:
> I think we are almost there! I made a few comments/suggestions, PTAL.
> And, of course, MOAR UNIT TESTS :)

TFTR, didn't add unit tests previously to avoid having to rework on them if anything needed change based on your review comments. Added them now.

proxy/src/main/java/com/wavefront/agent/preprocessor/predicate/ContainsPredicate.java
akodali18: done, Thanks for the detailed review comments.

PR #513, opened 2020-04-03
PUB-225: Java 11 migration

basilisk487:


panghy:
Is there a reason why we need to upgrade the proxy to Java 11?

basilisk487:
@panghy mostly to take advantage of compact strings which would help reducing the memory footprint and GC. We don't have to upgrade the target bytecode for that, ofc, and just bundle it with a newer JRE.

panghy:
Okay, should note that this is mostly a java 9 thing but java 11 is the lts JRE.

PR #514, opened 2020-04-09
v2preprocessor - Enable In predicate to take in a yaml list 

akodali18:
Modify In predicate to take in an yaml list instead of comma separated values.
Rename "noop" -> "ignore"

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/preprocessor/PreprocessorUtil.java
+++ b/proxy/src/main/java/com/wavefront/agent/preprocessor/PreprocessorUtil.java
@@ -262,34 +262,70 @@ public abstract class PreprocessorUtil {
     if (reportableEntity == ReportPoint.class) {
       switch (subElement.getKey()) {
         case "equals":
-          return new ReportPointEqualsPredicate(scope, (String) ruleVal);
+          if (ruleVal instanceof String) {
+            return new ReportPointEqualsPredicate(scope, (String) ruleVal);
+          }
+          throw new IllegalArgumentException("Argument [equals] should have a value of type [string].");
         case "startsWith":
-          return new ReportPointStartsWithPredicate(scope, (String) ruleVal);
+          if (ruleVal instanceof String) {
+            return new ReportPointStartsWithPredicate(scope, (String) ruleVal);
+          }
+          throw new IllegalArgumentException("Argument [startsWith] should have a value of type [string].");
         case "contains":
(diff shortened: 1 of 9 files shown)


proxy/src/main/java/com/wavefront/agent/preprocessor/PreprocessorUtil.java
```diff
@@ -261,34 +262,34 @@ private static Predicate processComparisonOp(Map.Entry<String, Object> subElemen
     if (reportableEntity == ReportPoint.class) {
       switch (subElement.getKey()) {
         case "equals":
-          return new ReportPointEqualsPredicate(scope, ruleVal);
+          return new ReportPointEqualsPredicate(scope, (String) ruleVal);
```
basilisk487, line 265: I think it would be a good idea to check whether ruleVal type is the expected one and throw an exception with a meaningful message, i.e. if "in" is used with String or "equals" used with List. Also add unit tests to verify that these situations are handled correctly.

proxy/src/main/java/com/wavefront/agent/preprocessor/PreprocessorUtil.java
```diff
@@ -261,34 +262,34 @@ private static Predicate processComparisonOp(Map.Entry<String, Object> subElemen
...
+          return new ReportPointContainsPredicate(scope, (String) ruleVal);
         case "endsWith":
-          return new ReportPointEndsWithPredicate(scope, ruleVal);
+          return new ReportPointEndsWithPredicate(scope, (String) ruleVal);
         case "regexMatch":
-          return new ReportPointRegexMatchPredicate(scope, ruleVal);
+          return new ReportPointRegexMatchPredicate(scope, (String) ruleVal);
         case "in":
-          return new ReportPointInPredicate(scope, ruleVal);
+          return new ReportPointInPredicate(scope, (ArrayList) ruleVal);
```
basilisk487, line 275: List<String>. In this context specific implementation (ArrayList) is not important but the element's type kind of is.

proxy/src/main/java/com/wavefront/agent/preprocessor/PreprocessorUtil.java
akodali18: Done.

akodali18:
Opened a new PR based on an alternative solution - https://github.com/wavefrontHQ/wavefront-proxy/pull/515

PR #515, opened 2020-04-09
Modify predicates to support both string and list values, rename noop…

akodali18:
… to ignore, added default preprocessor example.


proxy/src/main/java/com/wavefront/agent/preprocessor/predicate/ComparisonPredicate.java
```diff
@@ -9,10 +14,13 @@
...
+  protected final List<String> value;
 
-  public ComparisonPredicate(String scope, String value) {
+  public ComparisonPredicate(String scope, Object value) {
     this.scope = scope;
-    this.value = value;
+    Preconditions.checkArgument(value instanceof List || value instanceof String,
+        "Argument [value] should be of type [string] or [list].");
+    this.value =  (value instanceof List) ? ((List) value) :
+        Collections.unmodifiableList(Arrays.asList((String) value));
```
basilisk487, line 24: I know premature optimization is the root of all evil, but I'm still curious about performance implications of wrapping a single string with a List and using List.stream(), especially considering that most real-life use cases would probably have just one value. I would expect using stream() with a single-element list would be about 10-15 times slower than simply looping through the collection. Won't be as elegant, of course. 
We can leave this for later iterations; LGTM for now.

PR #516, opened 2020-04-11
Fix javadoc issues

basilisk487:


pom.xml
```diff
@@ -54,6 +54,7 @@
     <truth.version>0.29</truth.version>
 
     <java.version>1.8</java.version>
```
ajayj89, line 56: shouldn't we change this to 11 as well? or we want the compiler plugin to still use 1.8?

basilisk487: We are generating 1.8 binaries; we are just using jdk11 as runtime and for building

PR #518, opened 2020-04-16
Updated MetricsMatcher

howardyoo:
1. Prevent log message parsing to contain tags with empty values from being added to the telemetry datum.

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/config/MetricMatcher.java
+++ b/proxy/src/main/java/com/wavefront/agent/config/MetricMatcher.java
@@ -182,7 +182,7 @@ public class MetricMatcher extends Configuration {
     for (int i = 0; i < tagKeys.size(); i++) {
       String tagKey = tagKeys.get(i);
       if (tagValues.size() > 0) {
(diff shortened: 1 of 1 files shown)


proxy/src/main/java/com/wavefront/agent/config/MetricMatcher.java
```diff
@@ -182,7 +182,10 @@ public TimeSeries timeSeries(LogsMessage logsMessage, Double[] output) throws Nu
     for (int i = 0; i < tagKeys.size(); i++) {
       String tagKey = tagKeys.get(i);
       if (tagValues.size() > 0) {
-        tags.put(tagKey, expandTemplate(tagValues.get(i), matches));
+        String value = expandTemplate(tagValues.get(i), matches);
+        if(value.trim().length() > 0) {
```
basilisk487, line 186: StringUtils.isNotBlank()

howardyoo: Ok, updated

PR #522, opened 2020-04-22
PUB-226: add backend based configuration of tracing span sampling rate.

srujann:


proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -1024,6 +1027,10 @@ protected void processConfiguration(AgentConfiguration config) {
         entityProps.getGlobalProperties().
             setHistogramStorageAccuracy(config.getHistogramStorageAccuracy().shortValue());
       }
+      entityProps.getGlobalProperties().setTraceSamplingRate(config.getSpanSamplingRate());
+      rateSampler.setSamplingRate(entityProps.getGlobalProperties().getTraceSamplingRate());
+      logger.fine("Proxy trace span sampling rate set to " +
```
basilisk487, line 1032: Nit: maybe do a logger.info instead but only when the value actually changes? lgtm otherwise.

srujann: done.

PR #528, opened 2020-05-14
Support sampling/filtering of span logs using updated span log data f…

hanwavefront:
…ormat

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/listeners/tracing/TracePortUnificationHandler.java
+++ b/proxy/src/main/java/com/wavefront/agent/listeners/tracing/TracePortUnificationHandler.java
@@ -74,7 +74,9 @@ public class TracePortUnificationHandler extends AbstractLineDelimitedHandler {
   private final Supplier<Boolean> spanLogsDisabled;
 
   protected final Counter discardedSpans;
+  protected final Counter discardedSpanLogs;
   private final Counter discardedSpansBySampler;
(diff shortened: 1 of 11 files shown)


basilisk487:
LGTM (pending java-lib release)

proxy/src/main/java/com/wavefront/agent/listeners/tracing/TracePortUnificationHandler.java
```diff
@@ -199,6 +197,76 @@ public static void preprocessAndHandleSpan(
...
+              handler.reject(spanLogs, spanMessageHolder[0]);
+            } else {
+              handler.block(spanLogs);
+            }
+            return;
+          }
+        }
+        List<Span> spanOutput = new ArrayList<>(1);
+        try {
+          spanDecoder.decode(spanMessage, spanOutput, "dummy");
```
srujann, line 240: `dummy`?

basilisk487: it doesn't really matter in this context, but we should probably standardize on something like "default" or "unknown"

hanwavefront: yeah, right now i think we're defaulting to "dummy" in most places in the code. let me know if i should change that

PR #532, opened 2020-05-26
Original zipkin id as tag

hanwavefront:


proxy/src/main/java/com/wavefront/agent/listeners/tracing/ZipkinPortUnificationHandler.java
```diff
@@ -247,6 +247,10 @@ private void processZipkinSpan(zipkin2.Span zipkinSpan) {
     // Add application tags, span references, span kind and http uri, responses etc.
     List<Annotation> annotations = new ArrayList<>();
 
+    // Add original Zipkin trace and span ids as tags to make finding them easier
+    annotations.add(new Annotation("originalSpanId", zipkinSpan.id()));
```
akodali18, line 251: Overall change lgtm. 
But what is the use case for this, just for our debugging or some customer ask ? This would really blow up cardinality isn't it?

hanwavefront: Yeah, traceId was a customer ask, and figured we should add spanId as well.  I think we should be ok with the high cardinality as long as these tags don't get indexed.

akodali18: I will let @basilisk487 comments on this - i.e if this might cause any issues while querying with the unindexed tags as well, given the cardinality. We do have cases today where pod id's make point tags , but this would be many times that cardinality rate.

sushantdewan123:
@hanwavefront we should do this for Jaeger as well.
Also rename to zipkinSpanId instead of originalSpanId
and similarly jaegerSpanId instead of originalSpanId

hanwavefront:
@sushantdewan123 added for jaeger and renamed the zipkin tags

PR #535, opened 2020-05-26
PUB-203 Minor change to TLS configuration

srengasamy:


proxy/src/main/java/com/wavefront/agent/AbstractAgent.java
```diff
@@ -126,10 +126,13 @@ void initSslContext() throws SSLException {
       sslContext = SslContextBuilder.forServer(new File(proxyConfig.getPrivateCertPath()),
               new File(proxyConfig.getPrivateKeyPath())).build();
     }
-    if (proxyConfig.isEnableTLS()) {
+    if (!isEmpty(proxyConfig.getTlsPorts()) && sslContext == null) {
```
jbau, line 129: do you need a conditional based on secureAllPorts here?

srengasamy: This is **init** method. **secureAllPorts** will be derived from **tlsPorts** in the next line. This is just validation.

jbau: kk

PR #536, opened 2020-05-27
Support Jaeger Integration Protobuf format data via gRPC

haosong:


haosong:
I include the change of https://github.com/wavefrontHQ/wavefront-proxy/pull/526 since that one is not merged yet, but required in this PR as well.

proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerProtobufUtils.java
```diff
@@ -0,0 +1,361 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.function.Supplier;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+import static com.google.protobuf.util.Timestamps.*;
+import static com.google.protobuf.util.Durations.*;
```
hanwavefront, line 41: nit: use single class imports

proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerProtobufUtils.java
```diff
@@ -0,0 +1,361 @@
...
+    if (wfInternalReporter != null) {
+      // report converted metrics/histograms from the span
+      List<Pair<String, String>> spanTags = annotations.stream().map(a -> new Pair<>(a.getKey(),
+          a.getValue())).collect(Collectors.toList());
+      discoveredHeartbeatMetrics.add(reportWavefrontGeneratedData(wfInternalReporter,
+          span.getOperationName(), applicationName, serviceName, cluster, shard, sourceName,
+          componentTagValue, isError,
+          toMicros(span.getDuration()),
+          traceDerivedCustomTagKeys,
+          spanTags));
```
hanwavefront, line 321: nit: formatting

proxy/pom.xml
```diff
@@ -49,6 +49,11 @@
       <artifactId>jaeger-thrift</artifactId>
       <version>0.31.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.skywalking</groupId>
+      <artifactId>receiver-proto</artifactId>
+      <version>7.0.0</version>
+    </dependency>
```
hanwavefront, line 56: Can we include the jaeger-proto and grpc artifacts directly or does this library provide something on top that we need? Thinking that this library might add bloat and make it harder to maintain and update dependencies

haosong: Yeah I checked the `jaeger-proto` maven pkg, for some reason it doesn't include this https://github.com/jaegertracing/jaeger-idl/blob/master/proto/api_v2/collector.proto, so I have to switch to that one.

haosong: switched to opentelementry one

PR #537, opened 2020-05-29
Update dependencies

basilisk487:


proxy/src/main/java/com/wavefront/agent/ProxyCheckInScheduler.java
```diff
@@ -184,6 +184,9 @@ private AgentConfiguration checkin() {
             throw new RuntimeException("Aborting start-up");
           }
           break;
+        case 406:
+          //
+          break;
```
hanwavefront, line 189: is this intended?

proxy/src/main/java/com/wavefront/agent/ProxyCheckInScheduler.java
```diff
@@ -192,6 +195,9 @@ private AgentConfiguration checkin() {
             throw new RuntimeException("Aborting start-up");
           }
           break;
+        case 429:
+          //
+          break;
```
hanwavefront, line 200: same here, is this intended?

basilisk487: yup, we are ignoring 429s on purpose. updated the comment

PR #542, opened 2020-06-05
Report derived metrics with updated values post span is transformed with preprocessor rules.

akodali18:
Also fixed bug in CustomTracingPortHandler , where CLUSTER_TAG_KEY is wrongly assigned value of service key.

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerThriftUtils.java
+++ b/proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerThriftUtils.java
@@ -314,11 +314,37 @@ public abstract class JaegerThriftUtils {
         spanLogsHandler.report(spanLogs);
       }
     }
+
     // report stats irrespective of span sampling.
     if (wfInternalReporter != null) {
-      // report converted metrics/histograms from the span
-      List<Pair<String, String>> spanTags = annotations.stream().map(a -> new Pair<>(a.getKey(),
-          a.getValue())).collect(Collectors.toList());
+      // Set post preprocessor rule values and report converted metrics/histograms from the span
+      sourceName = wavefrontSpan.getSource();
+      List<Annotation> processedAnnotations = wavefrontSpan.getAnnotations();
+      for (Annotation processedAnnotation : processedAnnotations) {
+        switch (processedAnnotation.getKey()) {
+          case APPLICATION_TAG_KEY:
+            applicationName = processedAnnotation.getValue();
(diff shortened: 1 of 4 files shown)


sushantdewan123:
We should add a test case for the code that was changed in Jaeger and Zipkin integration.

proxy/src/main/java/com/wavefront/agent/listeners/tracing/JaegerThriftUtils.java
```diff
@@ -314,11 +314,37 @@ private static void processSpan(io.jaegertracing.thriftjava.Span span,
...
+          case ERROR_SPAN_TAG_KEY:
+            isError = processedAnnotation.getValue().equals(ERROR_SPAN_TAG_VAL);
+            continue;
+        }
+      }
+      List<Pair<String, String>> spanTags = processedAnnotations.stream().map(
+          a -> new Pair<>(a.getKey(), a.getValue())).collect(Collectors.toList());
+      // TODO: Modify to use new method from wavefront internal reporter.
       discoveredHeartbeatMetrics.add(reportWavefrontGeneratedData(wfInternalReporter,
           span.getOperationName(), applicationName, serviceName, cluster, shard, sourceName,
```
hanwavefront, line 349: use the post-processor `wavefrontSpan.getName()` instead of `span.getOperationName()`

proxy/src/main/java/com/wavefront/agent/listeners/tracing/ZipkinPortUnificationHandler.java
```diff
@@ -405,9 +405,33 @@ private void processZipkinSpan(zipkin2.Span zipkinSpan) {
...
+            continue;
+          case ERROR_SPAN_TAG_KEY:
+            isError = true;
+            continue;
+        }
+      }
+      List<Pair<String, String>> spanTags = processedAnnotations.stream().map(
+          a -> new Pair<>(a.getKey(), a.getValue())).collect(Collectors.toList());
       discoveredHeartbeatMetrics.add(reportWavefrontGeneratedData(wfInternalReporter,
           spanName, applicationName, serviceName, cluster, shard, sourceName, componentTagValue,
```
hanwavefront, line 436: use the post-processor `wavefrontSpan.getName()` instead of `spanName`

hanwavefront:
Perhaps not within the scope of this fix, but we should probably also update `isForceSampled` and `isDebugSpanTag` after preprocessor rules are applied.

PR #546, opened 2020-06-11
MONIT-19711, fix JAVA_HOME in installation script

ajayj89:
once committed, will need to cherry-pick to 7.x as well


ajayj89:
abandoned in favor of #547

PR #547, opened 2020-06-11
Improve reliability of pre-installed java version detection

basilisk487:


pkg/etc/init.d/wavefront-proxy
```diff
@@ -106,28 +107,36 @@ download_jre() {
...
+# locally installed JDK first. We will accept 8, 9, 10, 11.
 if [[ -z "$PROXY_JAVA_HOME" && ! -r $proxy_jre_dir/bin/java ]]; then
-    if type -p java; then
-        JAVA_VERSION=$(java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1)
-        if [[ $JAVA_VERSION == "11" || $JAVA_VERSION == "10" || $JAVA_VERSION == "9" || $JAVA_VERSION == "8" || $JAVA_VERSION == "1.8" ]]; then
-            JAVA_HOME=$($(dirname $(dirname $(readlink -f $(which javac)))))
-            echo "Using Java runtime $JAVA_VERSION detected in $JAVA_HOME" >&2
+    # if java found in path
+    if type -p java > /dev/null 2>&1; then
+        JAVA_VERSION=$(java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1) || true
```
ajayj89, line 122: i personally prefer awk :)
```java -version 2>&1 | head -1 | awk -F "\"" '{ print $2 }'

File: pkg/etc/init.d/wavefront-proxy [hunk 2]
```diff
@@ -106,28 +107,36 @@ download_jre() {
...
+        if [[ -z "$JAVA_VERSION" ]]; then
+          JAVA_VERSION="(unknown version)"
+        fi
+        if [[ $JAVA_VERSION == "11" || $JAVA_VERSION == "10" || $JAVA_VERSION == "9" || $JAVA_VERSION == "8" ]]; then
+            JAVA_HOME=$(java -XshowSettings:properties -version 2>&1 > /dev/null | grep 'java.home' | sed 's/^.* = //') || true
```
ajayj89, line 127: same here, awk, but sed is fine too. :)
```
java -XshowSettings:properties -version 2>&1 > /dev/null | grep 'java.home' | awk -F " = " '{ print $2 }'
```

pkg/etc/init.d/wavefront-proxy
```diff
@@ -106,28 +107,36 @@ download_jre() {
...
-        JAVA_VERSION=$(java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1)
-        if [[ $JAVA_VERSION == "11" || $JAVA_VERSION == "10" || $JAVA_VERSION == "9" || $JAVA_VERSION == "8" || $JAVA_VERSION == "1.8" ]]; then
-            JAVA_HOME=$($(dirname $(dirname $(readlink -f $(which javac)))))
-            echo "Using Java runtime $JAVA_VERSION detected in $JAVA_HOME" >&2
+    # if java found in path
+    if type -p java > /dev/null 2>&1; then
+        JAVA_VERSION=$(java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1) || true
+        if [[ -z "$JAVA_VERSION" ]]; then
+          JAVA_VERSION="(unknown version)"
+        fi
```
ajayj89, line 125: why do we need to set it to unknown? it will in any case be installed when the flow goes to line number #134 else block

basilisk487: purely for message output reason - "Found Java runtime (unknown version), needs 8, 9, 10 or 11 to run" reads better than "Found Java runtime , needs 8, 9, 10 or 11 to run"

pkg/etc/init.d/wavefront-proxy
basilisk487: that command doesn't quite work unfortunately:
``` 
$ java -version 2>&1 | head -1 | awk -F "\"" '{ print $2 }'
1.8.0_202-ea
$ java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1
8
```

pkg/etc/init.d/wavefront-proxy
basilisk487: Here it actually makes more sense, done

pkg/etc/init.d/wavefront-proxy
basilisk487: and you still need to remove the leading "1." in order to be able to split it by "." afterwards.

ajayj89: i see, we can also use `java.specification.version` from `java -XshowSettings:properties`, but i see we want `8` for `1.8` as well.

ajayj89:
minor comments, otherwise lgtm

ajayj89:
lgtm

PR #550, opened 2020-06-19
User-friendly preprocessor rules conditions 

basilisk487:
Missing unit tests and javadocs


proxy/src/main/java/com/wavefront/agent/preprocessor/predicate/MathExpression.java
```diff
@@ -0,0 +1,45 @@
...
+
+  public MathExpression(EvalExpression left, EvalExpression right, String op) {
+    this.left = left;
+    this.right = right;
+    this.op = op.replace(" ", "");
+  }
+
+  @Override
+  public double getValue(Object entity) {
+    switch (op) {
```
MikeMcMahon, line 18: call me crazy, but supporting bit operations might be useful here.  I've wanted bit wise operations for eons in wavefront so you can do true flags like 0000 0001 0010...etc.  I can't fathom a reason for having it here (unless you wanted to use bit operations to determine how to tag or sample a metric...buuuut it's a thought)

basilisk487: > do we have any examples of how one might use random() it looks like it might be for sampling a metric since we return a double?

Yup, that's for sampling, i.e.
if: 'random() > 0.75' gives you 25% sampling

basilisk487: > call me crazy, but supporting bit operations might be useful here.
fair enough. done.

MikeMcMahon:
lgtm

PR #552, opened 2020-06-23
PUB-236: Bursty traffic smoothing

basilisk487:


proxy/src/main/java/com/wavefront/agent/handlers/TrafficShapingRateLimitAdjuster.java
```diff
@@ -0,0 +1,73 @@
...
+import com.yammer.metrics.stats.Sample;
+
+/**
+ * Experimental: use automatic traffic shaping (set rate limiter based on last 5 minutes of
+ * received rates
+ *
+ * @author vasily@wavefront.com.
+ */
+public class TrafficShapingRateLimitAdjuster extends TimerTask implements Managed {
+  private static final int DEFAULT_SAMPLE_SIZE = 1028;
```
jbau, line 23: any reason for 1028 and not 1024?

proxy/src/main/java/com/wavefront/agent/handlers/TrafficShapingRateLimitAdjuster.java
```diff
@@ -0,0 +1,73 @@
...
+  private static final double DEFAULT_ALPHA = 0.015;
```
jbau, line 24: so basically this is chosen to be ~1/60?

proxy/src/main/java/com/wavefront/agent/handlers/TrafficShapingRateLimitAdjuster.java
```diff
@@ -0,0 +1,73 @@
...
+      if (rate > 0 || samples.get() > 0) {
+        samples.incrementAndGet();
+        Sample sample = perEntityStats.computeIfAbsent(type, x ->
+            new ExponentiallyDecayingSample(DEFAULT_SAMPLE_SIZE, DEFAULT_ALPHA));
+        sample.update(rate);
+        if (samples.get() >= 300) {
+          RecyclableRateLimiter rateLimiter = entityProps.get(type).getRateLimiter();
+          double suggestedLimit = (MIN_RATE_LIMIT + sample.getSnapshot().getMedian()) * HEADROOM;
+          if (Math.abs(rateLimiter.getRate() - suggestedLimit) > rateLimiter.getRate() * 0.05) {
+            rateLimiter.setRate(suggestedLimit);
```
jbau, line 57: does resetting the rate limit every second do weird things to the rate limiter?  (looked at RecyclableRateLimiter code and seems not, but though it should be asked.

Also do we want the median * 1.25?  or it seems like we may want a higher percentile measurement?

jbau:
also what happens when the rate limit hits?  does it queue or reject the point on the listener?

proxy/src/main/java/com/wavefront/agent/handlers/TrafficShapingRateLimitAdjuster.java
basilisk487: 1028 samples with 0.015 alpha seem to be the industry standard, which heavily biases the reservoir to the last 5 minutes. 

> also what happens when the rate limit hits? does it queue or reject the point on the listener?

Eventually we queue, but only after we exhaust in-memory buffers. I tried the "auto adjust the flush batch size" approach, that doesn't use the rate limiter, at first; but that didn't give us enough flexibility in managing traffic. We would have to dramatically increase the size of in-memory buffers, and even then, if(when) the buffer limit is breached, the entire contents of the buffer would be immediately drained to the queue. Which would in turn immediately trigger backlog processing threads so in the end it would basically cause an even larger burst of traffic, which would defeat the purpose.

proxy/src/main/java/com/wavefront/agent/handlers/TrafficShapingRateLimitAdjuster.java
basilisk487: Yes, I accounted for this kind of scenario early on - changing the limit all the time doesn't affect permits that have already been used up (so the timestamp for the next available permit won't change), but it does adjust the amount of "accumulated" unused permits accordingly.

basilisk487: I think you are right - we should go with something like p75 by default, and make that percentile configurable.

proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -1030,7 +1039,10 @@ protected void startHistogramListeners(List<String> ports,
           return (ReportableEntityHandler<T, U>) handlers.computeIfAbsent(handlerKey,
               k -> new HistogramAccumulationHandlerImpl(handlerKey, cachedAccumulator,
                   proxyConfig.getPushBlockedSamples(), granularity, validationConfiguration,
-                  granularity == null, blockedHistogramsLogger, VALID_HISTOGRAMS_LOGGER));
+                  granularity == null,
+                  rate -> entityProps.get(ReportableEntityType.POINT).
```
jbau, line 1043: POINT or HISTOGRAM here?

proxy/src/main/java/com/wavefront/agent/handlers/DeltaCounterAccumulationHandlerImpl.java
```diff
@@ -84,7 +92,9 @@ public DeltaCounterAccumulationHandlerImpl(
         aggregationIntervalSeconds, TimeUnit.SECONDS);
 
     String metricPrefix = handlerKey.toString();
-    this.reportedCounter = Metrics.newCounter(new MetricName(metricPrefix, "", "sent"));
+    MetricName reported = new MetricName(metricPrefix, "", "sent");
+    this.reportedCounter = Metrics.newCounter(reported);
+    this.reportedStats = new BurstRateTrackingCounter(reported, Metrics.defaultRegistry(), 1000);
```
jbau, line 97: what's actually updating this.reportStats?
when reportedCounter is incremented, does reportedStats also update, due to the shared MetricName ?

jbau: shouldn't the increment in reportAggregatedDeltaValue be changed to this.reportStats.inc()?

PR #569, opened 2020-09-10
Metric Tag Count for Points/Histograms/Spans

joannak0:
New metric in the proxy to find out how many point/histogram/span tags user is emitting

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/handlers/ReportPointHandlerImpl.java
+++ b/proxy/src/main/java/com/wavefront/agent/handlers/ReportPointHandlerImpl.java
@@ -34,6 +34,8 @@ class ReportPointHandlerImpl extends AbstractReportableEntityHandler<ReportPoint
   final ValidationConfiguration validationConfig;
   final Function<Histogram, Histogram> recompressor;
   final com.yammer.metrics.core.Histogram receivedPointLag;
+  final com.yammer.metrics.core.Histogram histogramTagCount;
+  final com.yammer.metrics.core.Histogram pointTagCount;
   final Supplier<Counter> discardedCounterSupplier;
 
   /**
@@ -67,6 +69,8 @@ class ReportPointHandlerImpl extends AbstractReportableEntityHandler<ReportPoint
     MetricsRegistry registry = setupMetrics ? Metrics.defaultRegistry() : LOCAL_REGISTRY;
     this.receivedPointLag = registry.newHistogram(new MetricName(handlerKey.toString() +
         ".received", "", "lag"), false);
+    this.pointTagCount = registry.newHistogram(new MetricName(handlerKey.toString(), "", "pointTagCount"), false);
+    this.histogramTagCount = registry.newHistogram(new MetricName(handlerKey.toString(), "", "histogramTagCount"), false);
     this.discardedCounterSupplier = Utils.lazySupplier(() ->
         Metrics.newCounter(new MetricName(handlerKey.toString(), "", "discarded")));
   }
@@ -84,6 +88,13 @@ class ReportPointHandlerImpl extends AbstractReportableEntityHandler<ReportPoint
       Histogram histogram = (Histogram) point.getValue();
       point.setValue(recompressor.apply(histogram));
     }
+
+    if (point.getValue() instanceof Histogram) {
+      histogramTagCount.update(point.getAnnotations().size());
+    } else {
+      pointTagCount.update(point.getAnnotations().size());
+    }
+
     final String strPoint = serializer.apply(point);
     getTask().add(strPoint);
     getReceivedCounter().inc();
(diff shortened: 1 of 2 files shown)


proxy/src/main/java/com/wavefront/agent/handlers/ReportPointHandlerImpl.java
```diff
@@ -67,6 +69,8 @@
     MetricsRegistry registry = setupMetrics ? Metrics.defaultRegistry() : LOCAL_REGISTRY;
     this.receivedPointLag = registry.newHistogram(new MetricName(handlerKey.toString() +
         ".received", "", "lag"), false);
+    this.pointTagCount = registry.newHistogram(new MetricName(handlerKey.toString(), "", "pointTagCount"), false);
```
sushantdewan123, line 72: rename this to 
pointTagCount.received

proxy/src/main/java/com/wavefront/agent/handlers/ReportPointHandlerImpl.java
```diff
@@ -67,6 +69,8 @@
...
+    this.histogramTagCount = registry.newHistogram(new MetricName(handlerKey.toString(), "", "histogramTagCount"), false);
```
sushantdewan123, line 73: rename this to 
histogramTagCount.received

proxy/src/main/java/com/wavefront/agent/handlers/SpanHandlerImpl.java
```diff
@@ -52,6 +56,8 @@
     this.validationConfig = validationConfig;
     this.validItemsLogger = validItemsLogger;
     this.dropSpansDelayedMinutes = dropSpansDelayedMinutes;
+    this.spanTagCount = Metrics.defaultRegistry().newHistogram(new MetricName(handlerKey.toString(), "",
+        "spanTagCount"), false);
```
sushantdewan123, line 60: rename this to 
spanTagCount.received

sushantdewan123 (changes requested):

proxy/src/main/java/com/wavefront/agent/handlers/ReportPointHandlerImpl.java
srujann: probable points.tagCount.received.

srujann: feel `points.tagCount.received` is slightly better

proxy/src/main/java/com/wavefront/agent/handlers/SpanHandlerImpl.java
```diff
@@ -52,6 +56,8 @@
     this.validationConfig = validationConfig;
     this.validItemsLogger = validItemsLogger;
     this.dropSpansDelayedMinutes = dropSpansDelayedMinutes;
+    this.spanTagCount = Metrics.defaultRegistry().newHistogram(new MetricName(handlerKey.toString(), "",
```
srujann, line 59: use the same pattern as `ReportPointHandlerImpl` to determine the registry to use.

srujann (changes requested):

proxy/src/main/java/com/wavefront/agent/handlers/ReportPointHandlerImpl.java
basilisk487: I have to disagree with both @srujann and @sushantdewan123 :) 
We should try to keep our naming at least somewhat consistent - I suggest naming it `.received.tagCount`, similar to `.received.lag` right above it. The final metric names would be like `~proxy.points.2878.received.tagCount.p99` etc.

proxy/src/main/java/com/wavefront/agent/handlers/ReportPointHandlerImpl.java
```diff
@@ -84,6 +88,13 @@ void reportInternal(ReportPoint point) {
       Histogram histogram = (Histogram) point.getValue();
       point.setValue(recompressor.apply(histogram));
     }
+
+    if (point.getValue() instanceof Histogram) {
+      histogramTagCount.update(point.getAnnotations().size());
```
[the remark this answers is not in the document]
srujann: sounds good.

proxy/src/main/java/com/wavefront/agent/handlers/ReportPointHandlerImpl.java
sushantdewan123: @basilisk487 the only difference here is that this metric is recorded pre-validation.
Hence I put in the keyword `raw` in there :)

sushantdewan123: Are you okay with :)
**~proxy.points.2878.received.raw.tagCount.p99**

basilisk487: what else would go under `received.raw.` metric tree? i would imagine `.received.tagCount` would be pre-validation, and if we need to, we could add `.sent.tagCount` for post-validation?

joannak0:
@basilisk487 , please merge if you approve :) Thanks!

basilisk487 (approved):

PR #572, opened 2020-10-01
Propagate all Jaeger process tags as span tags

### LINKED ISSUE #423: Support propagating all jaeger process tags(other than “hostname”) as `span tags`

thepeterstone:
This removes the filter for `ip` tags when forwarding Jaeger traces and fixes #423.


hanwavefront (approved):
LGTM!

PR #573, opened 2020-10-09
Reduce disk space allocation when queueing

basilisk487:
Instead of doubling the size of a buffer file every time we need to grow it, maintain a collection of smaller shards (128MB by default).


MikeMcMahon (approved):
lgtm

PR #589, opened 2020-12-04
Add proxy setting to ignore span head sampling config from backend

xshipeng:


proxy/src/main/java/com/wavefront/agent/PushAgent.java
```diff
@@ -1154,7 +1154,9 @@ protected void processConfiguration(AgentConfiguration config) {
             setHistogramStorageAccuracy(config.getHistogramStorageAccuracy().shortValue());
       }
       double previousSamplingRate = entityProps.getGlobalProperties().getTraceSamplingRate();
-      entityProps.getGlobalProperties().setTraceSamplingRate(config.getSpanSamplingRate());
+      if (!proxyConfig.isBackendSpanHeadSamplingPercentIgnored()) {
+        entityProps.getGlobalProperties().setTraceSamplingRate(config.getSpanSamplingRate());
+      }
       rateSampler.setSamplingRate(entityProps.getGlobalProperties().getTraceSamplingRate());
       if (previousSamplingRate != entityProps.getGlobalProperties().getTraceSamplingRate()) {
         logger.info("Proxy trace span sampling rate set to " +
```
srujann, line 1162: all of this logic should be under the flag to avoid ambiguity.

srujann (changes requested):
One minor comment.

srujann (approved):

PR #606, opened 2021-03-18
Create new docker file specific for WF-proxy release (filename: Dockerfile-for-release)

sbhakta-vmware:
When we create WF-Proxy / proxy-next docker images for release, we need to replace the original Dockerfile's "install wf-proxy" section with lines to install the proxy from generated .deb files for the release. Until now, we were using `sed` commands to delete and replace specific lines to achieve this. This was not a scalable solution. Hence, we decided to move forward with creating our own Dockerfile-for-release to achieve this.


alchen1218 (approved):
lgtm

PR #610, opened 2021-03-20
[Dockerfile-for-release]: Since we're using Photon, use tdnf to install rpm instead of using dpkg and deb

sbhakta-vmware:
Since we're moving to Photon OS, need to update Dockerfile-for-release to use tdnf and rpm to install wf-proxy instead of dpkg and deb


alchen1218 (approved):
LGTM

PR #612, opened 2021-03-24
[Dockerfile-for-release]: copy log4j.xml to /etc/wavefront/wavefront-proxy

sbhakta-vmware:
Add the line from original Dockerfile which copies log4j.xml to /etc/wavefront/wavefront-proxy


alchen1218 (approved):
lgtm

PR #614, opened 2021-03-26
Updating libraries and adding exclusions which pull in duplicate old libraries

formanojhr:
These will avoid security CVEs caught by scan and avoid those vulnerabilities.


hanwavefront (approved):
LGTM

PR #615, opened 2021-03-26
Downmerging security sve fixes for lib versions

formanojhr:
excluding guava from protubuf to avoid pulling in a bad version. 2. Adding exclusion of xstream dependency pulled.


formanojhr:
bad pull request closing

PR #617, opened 2021-04-19
NPE fix for FileBasedTaskQueue

formanojhr:
This will fix the null pointer exception being thrown for remove method.


proxy/src/main/java/com/wavefront/agent/queueing/FileBasedTaskQueue.java
```diff
@@ -83,9 +83,11 @@ public void remove() throws IOException {
       this.head = taskConverter.fromBytes(task);
     }
     queueFile.remove();
-    int weight = this.head.weight();
-    currentWeight.getAndUpdate(x -> x > weight ? x - weight : 0);
-    this.head = null;
+    if(this.head != null) {
```
hanwavefront, line 86: nit: space after if

hanwavefront (approved):
small nit, LGTM otherwise

PR #618, opened 2021-04-20
NPE fix for FileBasedTaskQueue (#617)

sbhakta-vmware:
* Lib version upgrades. Updating libs : thrift  0.14.1.   netty version to 4.1.60.Final. httpclient to 4.5.13. Removing xstream since it does not seem to be used in the compilation.

* 1. Excluding guava from protubuf to avoid pulling in a bad version. 2. Adding exclusion of xstream dependency pulled.

* doing a null check for weight before initializing weight

* removing unnecessary variables


formanojhr (approved):
Approved merge

PR #620, opened 2021-04-29
MONIT-23047 - Proxy Preprocessor Rules to Obfuscate Host, Tags, etc

laullon:


proxy/src/main/java/com/wavefront/agent/preprocessor/ReportPointObfuscateTransformer.java
```diff
@@ -60,29 +60,36 @@ public ReportPoint apply(@Nullable ReportPoint reportPoint) {
           String metric = reportPoint.getMetric();
           metric = encode(metric);
           reportPoint.setMetric(metric);
+          ruleMetrics.incrementRuleAppliedCounter();
           break;
+
         case "sourceName":
+        case "hostName":
           String source = reportPoint.getHost();
           source = encode(source);
```
formanojhr, line 69: Consider using a different var maybe calling it encSource for better readability.

proxy/src/main/java/com/wavefront/agent/preprocessor/ReportPointObfuscateTransformer.java
```diff
@@ -60,29 +60,36 @@ public ReportPoint apply(@Nullable ReportPoint reportPoint) {
           String metric = reportPoint.getMetric();
           metric = encode(metric);
           reportPoint.setMetric(metric);
+          ruleMetrics.incrementRuleAppliedCounter();
           break;
+
         case "sourceName":
```
formanojhr, line 66: Do we need this ? There is no special logic at this point.

proxy/src/main/java/com/wavefront/agent/preprocessor/ReportPointObfuscateTransformer.java
```diff
@@ -60,29 +60,36 @@ public ReportPoint apply(@Nullable ReportPoint reportPoint) {
...
-      e.printStackTrace();
+      throw new RuntimeException("Error running Obfuscate rule on '" + scope + "' scope", e);
     } finally {
       ruleMetrics.ruleEnd(startNanos);
     }
     return reportPoint;
   }
 
-  private String encode(String value) throws UnsupportedEncodingException, IllegalBlockSizeException, BadPaddingException {
+  private String encode(@Nonnull String value) throws UnsupportedEncodingException, IllegalBlockSizeException, BadPaddingException {
```
formanojhr, line 92: UnsupportedEncodingException it not thrown with this method

proxy/src/main/java/com/wavefront/agent/preprocessor/ReportPointObfuscateTransformer.java
```diff
@@ -60,29 +60,36 @@ public ReportPoint apply(@Nullable ReportPoint reportPoint) {
           String metric = reportPoint.getMetric();
           metric = encode(metric);
           reportPoint.setMetric(metric);
+          ruleMetrics.incrementRuleAppliedCounter();
           break;
+
         case "sourceName":
+        case "hostName":
```
formanojhr, line 67: It seems like based on existing rules sourceName is where this logic fits in. E.g. from docs. Also once sourceName is there used hostName does not have relevance. 
https://docs.wavefront.com/proxies_preprocessor_rules.html#point-altering-rules. We also need. to support matches parameter. Please look at processors like SpanBlockFilter
example:case "sourceName":
          if (compiledPattern.matcher(span.getSource()).matches()) {
            ruleMetrics.incrementRuleAppliedCounter();
            return false;

proxy/src/test/resources/com/wavefront/agent/preprocessor/preprocessor_rules.yaml
```diff
@@ -395,3 +395,15 @@
     action        : spanAddTag
     key           : multiPortSpanTagKey
     value         : multiSpanTagVal
+
+  ## obfuscate Tests
+  - rule    : example-obfuscation
+    action  : obfuscate
+    key     : 1234567890123456
+    scope   : metricName
```
formanojhr, line 403: It seems like metricName is a key word scope; so you would need matching parameters with regex pattern like
- rule    : example-force-lowercase
    action  : forceLowercase
    scope   : metricName
    match   : "^UPPERCASE.*$"

https://docs.wavefront.com/proxies_preprocessor_rules.html#forcelowercase

proxy/src/test/resources/com/wavefront/agent/preprocessor/preprocessor_rules.yaml
```diff
@@ -395,3 +395,15 @@
...
+
+  - rule    : example-obfuscation-2
+    action  : obfuscate
+    key     : 1234567890123456
+    scope   : hostName
```
formanojhr, line 408: It seems like sourceName is the standard keyword here for matching hosts.  and you would need a match parameter
Examples
  # block all spans with sourceName that starts with qa-service
###############################################################
  - rule    : example-span-block-qa-services
    action  : spanBlock
    scope   : sourceName
    match   : "qa-service.*"

formanojhr (commented):
Added some review comment German. Sorry it took this long

proxy/src/main/java/com/wavefront/agent/preprocessor/ReportPointObfuscateTransformer.java
```diff
@@ -0,0 +1,90 @@
...
+  private final Cipher cipher;
+
+
+  public ReportPointObfuscateTransformer(final String key,
+                                         final String scope,
+                                         final PreprocessorRuleMetrics ruleMetrics) {
+    Preconditions.checkNotNull(key, "[key] can't be null");
+    this.scope = Preconditions.checkNotNull(scope, "[scope] can't be null");
+    Preconditions.checkArgument(!key.isEmpty(), "[key] can't be blank");
+    Preconditions.checkArgument((key.length() == 16), "[key] have to be 16 characters");
```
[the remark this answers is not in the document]
laullon: we are using 128, so it support 16 or 32, anyway, string to obfuscate are too small, so is not important.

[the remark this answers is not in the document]
formanojhr: cool sure thought the requirement was AES 256. We are still in proto so we dont have to worry too much

proxy/src/test/java/com/wavefront/agent/preprocessor/AgentConfigurationTest.java
```diff
@@ -33,7 +33,7 @@ public void testLoadValidRules() {
     InputStream stream = PreprocessorRulesTest.class.getResourceAsStream("preprocessor_rules.yaml");
     config.loadFromStream(stream);
     Assert.assertEquals(0, config.totalInvalidRules);
-    Assert.assertEquals(60, config.totalValidRules);
+    Assert.assertEquals(61, config.totalValidRules);
```
formanojhr, line 36: It seems like this test is failing. 
java.lang.AssertionError: 
Expected :61
Actual   :62
<Click to see difference>


	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at com.wavefront.agent.preprocessor.AgentConfigurationTest.testLoadValidRules(AgentConfigurationTest.java:36)
	at

formanojhr (commented):
Just the test needs to be fixed I think otherwise everything else looks good.

formanojhr (approved):

formanojhr:
You should be able to merge now to master.

PR #628, opened 2021-06-28
ESO:3222: change gpg keyserver from hkp://keys.gnupg.net to keyserver.ubuntu.com

sbhakta-vmware:
https://wavefront.atlassian.net/browse/ESO-3222


ajayj89 (approved):
Lgtm

PR #640, opened 2021-07-16
[PUB-304] Use CACerts within the WF Proxy Docker Container

laullon:


formanojhr (approved):
lgtm. Would be good to wait for TMC integration team jgee to sign off.

PR #646, opened 2021-09-07
Revert 10.8 -> 10.9

laullon:


ajayj89 (changes requested):
we don't need this, 10.8 is already released to staging artifacts:
https://github.com/wavefrontHQ/wavefront-proxy/tree/wavefront-10.8
https://packagecloud.io/app/wavefront/proxy-next/search?q=10.8

PR #657, opened 2021-10-26
[PUB-321] Add support for the line delimited format ingestion for Events 

thepeterstone:
This PR changes the way events are reported to use the line-delimited format on the reporting API rather than the entity API.

In conjunction with upcoming system changes, this will enable high-throughput collection of events via the proxy.


thepeterstone:
Closing this for now, we can open a new PR if needed

PR #660, opened 2021-10-29
Revert Obfuscate Preprocessor Rules

laullon:


laullon:
This was done as POC for a customer, but we didn't get feedback, so reverting for now.

PR #662, opened 2021-11-04
[MONIT-25601] Enable multiplatform support for arm64 for AWS deployments

laullon:


sbhakta-vmware (approved):
Since we're using `adoptopenjdk:11-jre-hotspot` instead of `photon`, I hope we have done vulnerability scans for the same.

Also, we can update the README with more information about `make` targets, especially about how to use `make` to create multi-arch releases.
https://github.com/wavefrontHQ/wavefront-proxy/blob/MONIT-25601/docker/README.md

Otherwise lgtm.

PR #693, opened 2021-12-21
New build system

laullon:


Makefile
```diff
@@ -0,0 +1,66 @@
+TS = $(shell date +%Y%m%d%H%M%S)
```
sbhakta-vmware, line 1: we can benefit from making the TS more readable instead of a string of numbers.
for example, separate date and time with a "_"
```
$ date +%Y%m%d_%H%M%S
20220110_141848
```

Makefile
```diff
@@ -0,0 +1,66 @@
...
+
+VERSION = $(shell mvn -f proxy -q -Dexec.executable=echo -Dexec.args='$${project.version}' --non-recursive exec:exec)
+REVISION ?= ${TS}
+FULLVERSION = ${VERSION}${REVISION}
```
sbhakta-vmware, line 5: should there some separation between the version and revision? Maybe a "_"
```
FULLVERSION = ${VERSION}_${REVISION}
```

sbhakta-vmware (approved):
trivial naming suggestions. Otherwise LGTM

Makefile
sbhakta-vmware: Just realized that `TS` is used as a default value for `REVISION`, and `REVISION` can be passed as an env var to the makefile, so we can make this change in the Jenkins job directly. Please ignore this suggestion. :)

johncornish:
Looks good to us on K8PO!

PR #702, opened 2022-02-09
[MONIT-25884] Improve error message when buffer location has incorrec…

joannak-vmware:
…t permissions

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/queueing/TaskQueueFactoryImpl.java
+++ b/proxy/src/main/java/com/wavefront/agent/queueing/TaskQueueFactoryImpl.java
@@ -116,9 +116,8 @@ public class TaskQueueFactoryImpl implements TaskQueueFactory {
       // fail if tryLock() returns null (lock couldn't be acquired)
       Preconditions.checkNotNull(channel.tryLock());
     } catch (Exception e) {
-      logger.severe("WF-005: Error requesting exclusive access to the buffer " +
-          "lock file " + lockFileName + " - please make sure that no other processes " +
-          "access this file and restart the proxy");
+      logger.severe("WF-005: Error writing to the buffer lock file " + lockFileName +
+          " - please make sure write permissions are correct for this file path and restart the proxy");
       return new TaskQueueStub<>();
     }
     try {
--- a/proxy/src/main/java/com/wavefront/agent/queueing/TaskQueueFactoryImpl.java
+++ b/proxy/src/main/java/com/wavefront/agent/queueing/TaskQueueFactoryImpl.java
@@ -16,7 +16,6 @@ import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.nio.channels.FileChannel;
 import java.nio.channels.OverlappingFileLockException;
-import java.nio.file.DirectoryNotEmptyException;
 import java.nio.file.Files;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


proxy/src/main/java/com/wavefront/agent/queueing/TaskQueueFactoryImpl.java
```diff
@@ -110,14 +111,31 @@ public Long value() {
     try {
       File lockFile = new File(lockFileName);
       if (lockFile.exists()) {
-        Preconditions.checkArgument(true, lockFile.delete());
+        if (!lockFile.delete()) {
```
laullon, line 114: Use `Files.Delete()` it give more info a better exceptions.
https://docs.oracle.com/javase/tutorial/essential/io/delete.html

proxy/src/main/java/com/wavefront/agent/queueing/TaskQueueFactoryImpl.java
```diff
@@ -110,14 +111,31 @@ public Long value() {
...
+          "proxy");
+      return new TaskQueueStub<>();
+    } catch (OverlappingFileLockException e) {
+      logger.severe("Error requesting exclusive access to the buffer " +
+          "lock file " + lockFileName + " - please make sure that no other processes " +
+          "access this file and restart the proxy");
+      return new TaskQueueStub<>();
+    } catch (IOException e) {
+      logger.severe("Error requesting access to buffer lock file " + lockFileName + " Channel is " +
+          "closed or an I/O error has occurred - please restart the proxy");
```
laullon, line 138: Add the exception message to all log messages.

proxy/src/main/java/com/wavefront/agent/queueing/TaskQueueFactoryImpl.java
laullon: or maybe best to use [deleteIfExists(Path)](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#deleteIfExists-java.nio.file.Path-)

joannak-vmware:
There are currently no unit tests for this, will be adding to this ticket

proxy/src/main/java/com/wavefront/agent/queueing/TaskQueueFactoryImpl.java
```diff
@@ -110,15 +113,36 @@ public Long value() {
     try {
       File lockFile = new File(lockFileName);
       if (lockFile.exists()) {
-        Preconditions.checkArgument(true, lockFile.delete());
+        if (!Files.deleteIfExists(lockFile.toPath())) {
+          // no file with name found
+          throw new LogFilePathException();
```
laullon, line 118: We don't care if the file can't be deleted if not found, that means that there is no other proxy blocking the buffer, we only care if the file exists and can't be deleted, and that throw a exception. So can remove the `if`.

proxy/src/main/java/com/wavefront/agent/queueing/TaskQueueFactoryImpl.java
```diff
@@ -110,15 +113,36 @@ public Long value() {
...
-    } catch (Exception e) {
-      logger.severe("WF-005: Error requesting exclusive access to the buffer " +
+      if (channel.tryLock() == null) {
+        throw new OverlappingFileLockException();
+      }
+    } catch (LogFilePathException e){
+          logger.severe("Error finding buffer lock file " + lockFileName +
+          " - please make sure file path is correct and restart the proxy: " + e);
+      return new TaskQueueStub<>();
+    } catch (DirectoryNotEmptyException e) {
```
joannak-vmware, line 129: Deleting catch as path will never be Directory (we create lockFile as new File above)

joannak-vmware:
QA Testing:
Please test and verify the log messages are correct when:
1. When User running proxy does not have write permission
2. Have proxy running, and try to run a second proxy on same proxyid
3. Buffer path is set to path user does not have permission to write to

laullon (approved):

sbhakta-vmware (approved):
lgtm

PR #707, opened 2022-03-04
update makefile

sbhakta-vmware:


Makefile
```diff
@@ -23,7 +24,7 @@ jenkins: .info build-jar build-linux push-linux docker-multi-arch clean
 #####
 build-jar: .info
 	mvn -f proxy --batch-mode package 
-	cp proxy/target/proxy-${VERSION}-uber.jar ${out}
+	cp proxy/target/${ARTIFACT_ID}-${VERSION}-uber.jar ${out}
```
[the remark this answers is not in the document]
sbhakta-vmware: better to have the source of truth from pom.xml rather than have it hardcoded in Makefile.
Since we changed our artifactId from `wavefront` to `proxy` in pom.xml, all jenkins jobs need to be changed. Trying to avoid this if we change anything in the future. Better to maintain the code. Let me know what you think.

Makefile
```diff
@@ -1,20 +1,20 @@
-TS = $(shell date +%Y%m%d-%H%M%S)
+TS := $(shell date +%Y%m%d-%H%M%S)
 
-VERSION = $(shell mvn -f proxy -q -Dexec.executable=echo -Dexec.args='$${project.version}' --non-recursive exec:exec)
+VERSION := $(shell mvn -f proxy -q -Dexec.executable=echo -Dexec.args='$${project.version}' --non-recursive exec:exec)
+ARTIFACT_ID := $(shell mvn -f proxy -q -Dexec.executable=echo -Dexec.args='$${project.artifactId}' --non-recursive exec:exec)
```
laullon, line 4: this take around 5 sec each... I may change it in a future, but is ok for now.

laullon (approved):
LGTM

Makefile
sbhakta-vmware: Sounds good. Thanks @laullon

joannak-vmware (approved):

PR #726, opened 2022-04-21
[MONIT-28397] Discrepancy in buffer file data and proxy backlog charts

laullon:


proxy/src/main/java/com/wavefront/agent/queueing/QueueController.java
```diff
@@ -51,10 +47,10 @@
   @SuppressWarnings("UnstableApiUsage")
   protected final RateLimiter reportRateLimiter = RateLimiter.create(0.1);
 
-  private AtomicLong currentWeight = null;
-  private final AtomicInteger queueSize = new AtomicInteger();
+  private long currentWeight;
+  private int queueSize;
```
jbau, line 51: do we think it's safe to remove the atomic?

jbau (commented):
would like to briefly discuss why it's safe to remove atomic, otherwise i think it looks ok.

proxy/src/main/java/com/wavefront/agent/queueing/QueueController.java
laullon: Yes, we create just one instance of this class for each "port", they are private members, update only on thread.
I don't see the need for Atomic operations.

laullon: but can be put Atomic back if you think is best

jbau: The metric reporting runs in a different thread, actually.  Although that thread does is read the int.  So if only one thread is writing we are still ok.  I'm fine keeping it int.

jbau (approved):

PR #727, opened 2022-04-21
[MONIT-28280] Add new metrics related with the Memory buffer

laullon:


proxy/src/main/java/com/wavefront/agent/handlers/AbstractSenderTask.java
```diff
@@ -55,6 +55,7 @@
   final Counter blockedCounter;
   final Counter bufferFlushCounter;
   final Counter bufferCompletedFlushCounter;
+  private final Histogram metricSize;
```
amitw-vmware, line 58: [future] we may want to call it atomSizes (or metricSizes in the future) but we aren't consistent about the term atom or metric so this works :)

amitw-vmware (approved):
LGTM

proxy/src/main/java/com/wavefront/agent/handlers/AbstractSenderTask.java
laullon: sure

joannak-vmware (approved):
LGTM

PR #744, opened 2022-05-11
Upgrade OTEL dependencies.

keep94:


proxy/pom.xml
```diff
@@ -619,7 +619,7 @@
     <dependency>
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-proto</artifactId>
-      <version>${opentelemetry.version}</version>
+      <version>1.7.1-alpha</version>
```
oppegard, line 622: If you [search maven central](https://search.maven.org/search?q=opentelemetry-proto) for the `opentelemetry-proto` artifact used in our pom.xml, you'll see that 1.7.1-alpha was released in Nov 2021. The other dependency on maven central is the replacement for that old, deprecated artifact. The replacement is currently 0.17.0-alpha and was release in May 2022.

oppegard (changes requested):

proxy/pom.xml
laullon: can we use "${opentelemetry.version}" on all dependencies ?

laullon:
Do we have a jira ticket for this ?

PR #748, opened 2022-05-20
WFESO-4697: enable Github workflow to notarize mac tarball proxy releases

sbhakta-vmware:


sbhakta-vmware:
No need to merge this on release-11.x. we need to just have it on `dev` branch

PR #656, opened 2021-10-25
OpenTelemetry HTTP & gRPC Listeners for Distributed Tracing

oppegard:
### Description
This feature enables the proxy to receive OpenTelemetry traces (via protobufs) over gRPC or HTTP listeners.

- The listeners are enabled with the new config options `otlpGrpcListenerPorts` and `otlpHttpListenerPorts`. They are not enabled by default.
- While the listeners only support trace data now, work is underway to enable them to receive OpenTelemetry metric data (via protobufs). This is the reasoning for not using the package namespace `com.wavefront.agent.listeners.tracing`, since these listeners won't be restricted to just traces.
- It also contains some basic refactoring within `PushAgent.java`.

### Testing Notes

To test this in an end-to-end manner, we've been using the [opentelemetry java agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation#getting-started) with the [spring petclinic sample app](https://github.com/spring-projects/spring-petclinic).

First, build this branch of the proxy and enable the gRPC listener with `otlpGrpcListenerPorts=4317`.

After building the petclinic repo and downloading [opentelemetry-javaagent.jar](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v1.11.1/opentelemetry-javaagent.jar), run it with the following:

```sh
# Assumes proxy is listening at localhost:4317
java -javaagent:/tmp/opentelemetry-javaagent.jar \
  -Dotel.service.name=petclinic \
  -Dotel.resource.attributes=application=pet-app  \
  -jar target/spring-petclinic*.jar
```
The above will result in WF traces with tags of `application=pet-app` and `service=petclinic`. At startup it will generate some traces related to in-memory DB migrations. After, you can visit localhost:8080 to generate traces demonstrating HTTP request/response work.

If testing the HTTP listener is desired, configure the proxy to run with `otlpHttpListenerPorts=4318` and run the petclinic app in the same way, but add the additional parameter of `-Dotel.exporter.otlp.protocol=http/protobuf`.

If verbose proxy logging is desired to view the ingested OTel structures and the resulting transformation into WF spans and span logs, the following config can be added to log4j:

```
    <Loggers>
        <Root level="INFO">
            <AppenderRef ref="Console" />
        </Root>
        <AsyncLogger name="OTLPDataLogger" level="ALL" additivity="false">
            <AppenderRef ref="Console"/>
       </AsyncLogger>
    </Loggers>
```

PR #760, opened 2022-06-08
[MONIT-28156] OpenTelemetry Metrics ingestion

oppegard:
### Description

This feature enables the proxy to receive OpenTelemetry Metric protobufs over the existing gRPC and HTTP listeners introduced in #656.

### Testing Notes

TBD – discuss with proxy team about available options.

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/listeners/otlp/OtlpHttpHandler.java
+++ b/proxy/src/main/java/com/wavefront/agent/listeners/otlp/OtlpHttpHandler.java
@@ -4,7 +4,6 @@ import com.google.common.collect.Sets;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.rpc.Code;
 import com.google.rpc.Status;
-
 import com.wavefront.agent.auth.TokenAuthenticator;
 import com.wavefront.agent.channel.HealthCheckManager;
 import com.wavefront.agent.handlers.HandlerKey;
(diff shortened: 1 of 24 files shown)


proxy/src/main/java/com/wavefront/agent/listeners/otlp/OtlpHttpHandler.java
```diff
@@ -129,18 +139,26 @@ protected void handleHttpMessage(ChannelHandlerContext ctx, FullHttpRequest requ
...
         default:
-          String msg = "Unknown OTLP endpoint " + uri.getPath();
-          logWarning("WF-300: " + msg, null, ctx);
-          HttpResponse response = makeErrorResponse(Code.NOT_FOUND, msg);
-          writeHttpResponse(ctx, response, request);
+          /*
+          We use HTTP 200 for success and HTTP 400 for errors, mirroring what we found in
+          OTel Collector's OTLP Receiver code.
+         */
+          // TODO report internal metrics
```
joannak-vmware, line 160: Has this `TODO` comment been addressed?

proxy/src/main/java/com/wavefront/agent/listeners/otlp/OtlpMetricsUtils.java
```diff
@@ -0,0 +1,580 @@
...
+                                       @Nullable Supplier<ReportableEntityPreprocessor> preprocessorSupplier,
+                                       String defaultSource,
+                                       boolean includeResourceAttrsForMetrics) {
+    ReportableEntityPreprocessor preprocessor = null;
+    if (preprocessorSupplier != null) {
+      preprocessor = preprocessorSupplier.get();
+    }
+
+    for (ReportPoint point : fromOtlpRequest(request, preprocessor, defaultSource, includeResourceAttrsForMetrics)) {
+      // TODO: handle sampler
```
joannak-vmware, line 68: Another `todo` comment, does this need to be addressed in this PR?

keep94: Removed as there is no sampler for metrics/points.

proxy/src/main/java/com/wavefront/agent/listeners/otlp/OtlpHttpHandler.java
keep94: Are there existing templates of internal metrics that other listeners report (e.g the listener on port 2878)? We don't want to invent new metrics that may not have value.

joannak-vmware: These are the existing internal metrics we currently have for proxy: https://docs.wavefront.com/monitoring_proxies.html#monitoring-proxy-health-with-the-usage-dashboard. We also have internal metrics for errors in system/CollectorApiServer.

keep94: Thanks for the info Joanna. Looking at other parts of the WF Proxy code, it appears that updating internal metrics upon sending 400 http status is not done. After consulting with my team, we feel that we don't want to add special internal metrics for 400 statuses just for OTEL.  If later we want to start an internal metric for 400 status codes, it should be a separate story and apply to all intake formats and not be a part of this PR.

joannak-vmware (approved):
LGTM

sbhakta-vmware (approved):
lgtm. tested the functionality with @deosu

PR #761, opened 2022-06-10
WFESO-4697: fix mac tarball workflow

sbhakta-vmware:
Remove `echo` a `sleep` introduced during testing in github workflow for mac tarball notarization


joannak-vmware (approved):
LGTM

laullon (approved):

PR #774, opened 2022-07-22
[MONIT-29838] ExtractTag and extractTagIfNotExists Preprocessor Update

joannak-vmware:
This commit includes point tag key in search, as source=pointLine rule applies to the whole point line.
Testing:

Add the following to proxy/preprocessor_rules.yaml:
```
  # add tag hasTenantId
  - rule : extract-tenantfinal
    action : extractTagIfNotExists
    source : pointLine
    tag : newAddedTag
    search : ".*device.*"
    replace : "newTag"
```

Confirm that new tag newAddedTag is added to metric with value newTag when **point tag key** matches search regex.
When action: extractTag and tag newAddedTag already exists, confirm new value is set.
When action: extractTagIfNotExists and tag newAddedTag already exists, confirm original value is kept.


laullon (approved):
LGTM

sbhakta-vmware (approved):
lgtm

PR #775, opened 2022-07-22
[MONIT-28882] Use a non named user

laullon:
Use a non named user to enable proxy deployment on security enabled k8s clusters like TKGS or TMC with TKGm


joannak-vmware (approved):
LGTM

akodali18:
@laullon: Thanks for the PR. This is no longer a requirement from k8s side, so please feel free to close this without merge. We can probably keep using a named user for the proxy container as its a more friendlier way to provide a user/debug for the operator.

PR #778, opened 2022-08-02
MONIT-29906: Span log sampling synchronization between proxy and DI -…

zen10001:
Span log is a separate request than span in WF. On DI side, it does not know if a span log has already been sampled or not as it can come from direct ingestion. It currently uses span field in span log for sampling and filtering. We need to sync span log sampling between proxy and DI and fix the cases that it escape the span log sampling at all.


proxy/src/main/java/com/wavefront/agent/listeners/otlp/OtlpTraceUtils.java
```diff
@@ -140,6 +148,15 @@ public static void exportToWavefront(
     }
   }
 
+  private static void addSpanLine(Span span, SpanLogs spanLogs) {
```
deosu, line 151: Does it make sense to move this piece of code to some Util class?

deosu (approved):
Just a small comment, otherwise LGTM.

proxy/src/main/java/com/wavefront/agent/listeners/tracing/SpanUtils.java
```diff
@@ -185,4 +184,13 @@ public static String toStringId(ByteString id) {
...
     return uuid.toString();
   }
+
+  public static void addSpanLine(Span span, SpanLogs spanLogs) {
+    String policyId = null;
+    if (span != null && span.getAnnotations() != null) {
+      policyId = AnnotationUtils.getValue(span.getAnnotations(),
+          SPAN_SAMPLING_POLICY_TAG);
+    }
+    spanLogs.setSpan(SPAN_SAMPLING_POLICY_TAG + "=" + (policyId == null ? "NONE" : policyId));
```
oppegard, line 194: Could we get an explicit unit test in `SpanUtilsTest` for this line? To see if any tests would fail I changed the line to:

`spanLogs.setSpan(SPAN_SAMPLING_POLICY_TAG + "=" + (policyId == null ? "NONE" : "FOO"));`

One test failed:

`com.wavefront.agent.listeners.tracing.JaegerTChannelCollectorHandlerTest.testJaegerDebugOverride`.

But, that seems to be a happy accident that a failure occurred, and isn't explicitly testing how `addSpanLine()` is supposed to behave.

oppegard (changes requested):
Just one small request for a bit of additional unit test coverage. Otherwise, lgtm.

proxy/src/main/java/com/wavefront/agent/listeners/tracing/SpanUtils.java
zen10001: There is a test case hitting the case when policyId is not null. It verifies that call.
https://github.com/zen10001/wavefront-proxy/blob/master/proxy/src/test/java/com/wavefront/agent/listeners/tracing/JaegerTChannelCollectorHandlerTest.java#:~:text=_sampledByPolicy%3Dtest
Is this sufficient?

deosu (approved):
LGTM

proxy/src/main/java/com/wavefront/agent/listeners/tracing/SpanUtils.java
oppegard: My only nit is that our single test case for when policyId is not null exists in `JaegerTChannelCollectorHandlerTest`. If that test is deleted (say Jaeger's Thrift over UDP is deprecated/deleted one day), then we don't have coverage for the other places it's used.

oppegard (approved):

deosu (approved):
LGTM

deosu (approved):

deosu (approved):

deosu (approved):

deosu (approved):

deosu (approved):

joannak-vmware (approved):
LGTM

PR #779, opened 2022-08-05
Accept OpenTelemetry metrics with int/long values

oppegard:
Prior to this fix, we assumed data points in OpenTelemetry metrics always had double values. If a data point instead had an int/long value, the resulting WF metric would have a likely incorrect value of 0.0.


deosu (approved):

joannak-vmware (approved):
LGTM

amitw-vmware (approved):
LGTM - approving to unblock the write

PR #785, opened 2022-08-23
Fix macos notarization according to new packaging

sbhakta-vmware:
based on this PR - https://github.com/wavefrontHQ/wavefront-proxy/pull/753
and https://jira.eng.vmware.com/browse/MONIT-27734


joannak-vmware (approved):
LGTM

PR #792, opened 2022-09-09
MONIT-30608: Update Proxy Retry to not retry if daily limit is reached

xuranchen:


amitw-vmware (approved):
LGTM

proxy/src/main/java/com/wavefront/agent/data/LogDataSubmissionTask.java
```diff
@@ -65,6 +67,14 @@ Response doExecute() {
     return api.proxyLogs(AGENT_PREFIX + proxyId.toString(), logs);
   }
 
+  @Override
+  protected TaskResult handleStatus429() {
+    Metrics.newCounter(
+            new MetricName(entityType + "." + handle, "", "failed" + ".ingestion_limit_reached"))
+        .inc(this.weight());
+    return TaskResult.REMOVED;
```
laullon, line 75: Is ok to drop logs on a 429 ? Why not retry ?

iprotsiuk (approved):
tested with docker and local proxy, lgtm

proxy/src/main/java/com/wavefront/agent/data/LogDataSubmissionTask.java
xuranchen: A 429 from VRLIC means that the daily ingestion limit has been reached

laullon: So, that's not coming from wavefront ?

xuranchen: Correct, its coming from the VRLIC backend

joannak-vmware (approved):

PR #813, opened 2022-11-21
MONIT-31965 Add debug level payload logging

deosu:
Add a logger in LogDataSubmissionTask and log a payload before it is sent to the backend(if the root logger's level is defined as DEBUG).

### DIFF
--- a/proxy/src/main/java/com/wavefront/agent/data/LogDataSubmissionTask.java
+++ b/proxy/src/main/java/com/wavefront/agent/data/LogDataSubmissionTask.java
@@ -17,6 +17,8 @@ import java.util.function.Supplier;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.ws.rs.core.Response;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 /**
  * A {@link DataSubmissionTask} that handles log payloads.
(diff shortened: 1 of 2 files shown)


xuranchen (approved):

proxy/src/main/java/com/wavefront/agent/data/LogDataSubmissionTask.java
```diff
@@ -26,6 +28,7 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "__CLASS")
 public class LogDataSubmissionTask extends AbstractDataSubmissionTask<LogDataSubmissionTask> {
+  private static final Logger LOGGER = LogManager.getLogger(LogDataSubmissionTask.class);
```
laullon, line 31: Create a logger just for this.
https://github.com/wavefrontHQ/wavefront-proxy/blob/825e17e1c34f702656c2c9b7e1b429c313490a38/proxy/src/main/java/com/wavefront/agent/listeners/otlp/OtlpTraceUtils.java#L167

deosu: @laullon Can you please explain a little bit? Are you suggesting creating a logger just for the `LogDataSubmissionTask.execute()`? Is the `OTLP_DATA_LOGGER` just an example? If yes, I have another question. The `LogDataSubmissionTask.java` class does not have any other functionality like the `OtlpTraceUtils.java` class. Does it make sense to create a logger just for the `execute()` method? Please guide.

deosu: @laullon please take a look again.

laullon (approved):

joannak-vmware (approved):
LGTM

PR #825, opened 2023-01-13
Fix for CVE-2022-3171

joannak-vmware:


locke-chappel (commented):
The protobuf related CVEs are resolved with this change. Looks good from the security side.

sbhakta-vmware (approved):
lgtm

PR #838, opened 2023-04-04
MONIT-34093 - Mitigation for CVE-2022-1471

ajackson:
- explicitly add dependency on snakeyaml since we import it directly in the code
- use SafeConstructor form of the Yaml object construction to mitigate CVE-2022-1471


ajackson:
Note the Dependency Review will fail due to explicit inclusion of snakeyaml 1.33 but the code mitigation using the `SafeConstructor` is applied. 
https://wiki.folio.org/display/SEC/SnakeYaml+SafeConstructor

PR #839, opened 2023-04-04
MONIT-34093 - Upgrade to snakeyaml 2.0 for CVE-2022-1471

ajackson:
- explicitly add dependency on snakeyaml since we import it directly in the code
- update test to use explict loadOptions on the YamlFactory for the objectmapper

### DIFF
--- a/proxy/pom.xml
+++ b/proxy/pom.xml
@@ -593,6 +593,11 @@
       <artifactId>proto-google-common-protos</artifactId>
       <version>2.0.1</version>
     </dependency>
+    <dependency>
+      <groupId>org.yaml</groupId>
+      <artifactId>snakeyaml</artifactId>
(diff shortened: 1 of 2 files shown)


ajackson:
All test pass but not sure what if anything this breaks in jackson-databind itself with snakeyaml 2.0

locke-chappel:
Assuming jackson-dataformat-yaml is compatible with SnakeYaml 2.x this is the preferred solution (they claim to use the parser directly so they may have some issues like we did in our tests).

This mitigates the CVE and addresses scanning issues.

laullon (approved):
LGTM

locke-chappel (approved):
Change resolves the CVE and based on jackson project issue tracker it should be compatible with the current version.

PR #844, opened 2023-04-14
MONIT-30797: CVE-2021-20293 (6.1) - wavefront-proxy (resteasy-jaxrs)

Norayr25:


proxy/pom.xml
```diff
@@ -381,6 +381,11 @@
         <artifactId>okhttp</artifactId>
         <version>4.9.3</version>
       </dependency>
+      <dependency>
+        <groupId>commons-io</groupId>
+        <artifactId>commons-io</artifactId>
+        <version>2.11.0</version>
```
laullon, line 387: Why?

proxy/pom.xml
```diff
@@ -598,6 +603,16 @@
       <artifactId>snakeyaml</artifactId>
       <version>2.0</version>
     </dependency>
+    <dependency>
+      <groupId>net.jcip</groupId>
```
laullon, line 607: why?

laullon (commented):
You need to solve the uint test fails

proxy/pom.xml
Norayr25: It was automatically added to recognize **NotThreadSafe** annotation.  Removed, thanks

proxy/pom.xml
Norayr25: > Why?

Removed

locke-chappel (commented):
resteasy-jaxrs is no longer detected. No new CVEs appear to have been introduced by using version 5.0.6.Final. Looks good from the security side.

laullon (approved):
LGTM

PR #857, opened 2023-06-26
MONIT-34528: wavefront-proxy (spring-boot-starter-log4j2).

DavidAzatyan:


locke-chappel:
I think this was fixed in the proxy 13.0 release - please check with the latest proxy code to make sure we aren't double fixing issues.

PR #865, opened 2023-08-16
MONIT-40976: Hide Proxy UI tokens in 13.0

Norayr25:
When requesting the proxy configuration file using the Wavefront UI or an API call, the responce always includes the token value. We'll now cover it up and only display the final four digits. For cspAppId, cspAppSecret, and cspAPIToken, the same applies.


joannak-vmware (approved):
LGTM

PR #873, opened 2023-09-14
[MONIT-40972 ] lz4-pure-java

laullon:


proxy/pom.xml
```diff
@@ -381,6 +381,11 @@
         <artifactId>okhttp</artifactId>
         <version>4.9.3</version>
       </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-compress</artifactId>
+            <version>1.24.0</version>
+        </dependency>
```
locke-chappel, line 388: Isn't commons-compress already included on line 334 above?

locke-chappel:
Overall, looks good from the security side - old lz4 library is removed and implementation now uses commons-compress.

proxy/pom.xml
laullon: yeah... quick test ;-)

locke-chappel: looks good now :)

PR #886, opened 2023-11-15
fix: bump io.grpc.grpc-bom to 1.53.0 to address  CVE-2023-1428

ajayk:


ajayk:
@joannak-vmware @sbhakta-vmware :  for review

